PR #1, opened 2018-07-30
Chore/run server easily

MatiasOS:
I was try to run the server with old URI `mongodb://${process.env.user}:${encodeURIComponent(process.env.password)}@${process.env.cluster}/${process.env.dbName}?ssl=true&replicaSet=${process.env.replicaSetName}&authSource=admin` 
and it is excessive secure and hard to configure.  
Dev env must be easy to run and start. 
I have replaced with  a single `env` variable `uri = mongodb://${process.env.MONGO_URI}`. 
It allow to run mongoDB local without configurations. 
Also, I have added a simple `README.md` file with the instructions.


MatiasOS:
I think that is a good idea to have a `develop` branch to separate the work in progress from work done.

SowmyaManohar:
I highly recommend working in a separate branch, as the code right here is actually not a development version.

MatiasOS:
@SowmyaManohar What do you mean with a separate branch?

SowmyaManohar:
The current Master branch is hosted in Azure. So it's kind of production branch. We can create a new branch if you want to work separately in HGNRest.

MatiasOS:
@SowmyaManohar That will be cool if you can add a branch to add my reviewed code.

PR #10, opened 2020-06-11
Feature/timer

fredcat:
Add Timer API. Clean some warnings.


chrisweilacker (approved):
Looks good tested it with your branch and see no problems.

PR #15, opened 2020-06-22
Feature/wbs4.2.4/auto re activate user

AndonIgnatov:
### Auto re-activate user feature.

1. In this implementation I've setup the cronjob to run 1 minute after midnight every day. The reActivateUser function will get logged in Sentry as well, actually should do 1 log when the job start and then one log for each user that was re-activated.
2. Fixed a couple of deprecation warnings on the server.

### DIFF
--- a/src/cronjobs/userProfileJobs.js
+++ b/src/cronjobs/userProfileJobs.js
@@ -19,8 +19,17 @@ const userProfileScheduledJobs = function () {
     'America/Los_Angeles',
   );
 
+  const updateUserStatusToActive = new CronJob(
+    '1 0 * * *', // Run this every day, 1 minute after mimdnight (PST).
+    userhelper.reActivateUser,
+    null,
+    false,
+    'America/Los_Angeles',
+  );
+
   assignBlueBadge.start();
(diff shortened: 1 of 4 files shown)


src/cronjobs/userProfileJobs.js
```diff
@@ -19,8 +19,17 @@ const userProfileScheduledJobs = function () {
     'America/Los_Angeles',
   );
 
+  const updateUserStatusToActive = new CronJob(
+    '1 0 * * *', // Run this every day, 1 minute after mimdnight (PST).
+    userhelper.reActivateUser,
+    null,
+    false,
+    'America/Los_Angeles',
```
getnithesh, line 27: Can we keep the time zone string somewhere as a generic constant so that any future zone changes would be a in a single place.

AndonIgnatov: Good point, we certainly could change it to a constant, however I don't think it's a big deal to keep those timezone strings as they are. I did discuss with Jae a few weeks ago the various possibilities of handling timezones and he wants it all in LA time (PST), so I don't think it will be changing.

PR #19, opened 2020-08-04
Feature/wbs1.6/summary section

AndonIgnatov:
Please review and merge the "feature/wbs1.6/summary section" branch. I've made sure to sync my branch with the latest "development branch so there won't be any conflicts.


src/helpers/reporthelper.js
```diff
@@ -0,0 +1,127 @@
...
+  /**
+   * Checks whether a date belongs to a specific week based on week index.
+   *
+   * @param {string} dueDate The date to check.
+   * @param {integer} weekIndex An integer in the range 0 - 2, where 0 (this week),
+   *                      1 (last week) or 2 (the week before last).
+   * @return {boolean} True if match, false otherwise.
+   */
+  const doesDateBelongToWeek = function (dueDate, weekIndex) {
+    const pstStartOfWeek = moment().tz('America/Los_Angeles').startOf('week').subtract(weekIndex, 'week');
```
AGCapdeville, line 59: Should we be setting the time zone for LA? Are we just going off California time, for now, just unsure if this should be hardcoded here? Might be a good idea to have a ‘const timeZone’ holding the time zone that we can alter later?

src/cronjobs/userProfileJobs.js
```diff
@@ -5,7 +5,15 @@ const userhelper = require('../helpers/userhelper')();
 const userProfileScheduledJobs = function () {
   const assignBlueBadge = new CronJob(
     '0 0 * * 0',
-    userhelper.assignBlueBadgeforTimeNotMet,
+    userhelper.assignBlueBadgesForTimeNotMetOrSummaries,
+    null,
+    false,
+    'America/Los_Angeles',
```
AGCapdeville, line 11:
```suggestion
    'America/Los_Angeles',
```
Here as well, will this need to be altered?

src/helpers/reporthelper.js
AndonIgnatov: Good question and yes it should be set to LA, I had that discussion with Jae already at some point and he wants it all LA time, so we could have it hard coded like this.

src/cronjobs/userProfileJobs.js
AndonIgnatov: Yes, this is actually crucial, it's important that it is PST here, again Jae's decision not mine.

AndonIgnatov:
Adam, regarding the UI component alignment issue, I saw that too in your video. Could you check that you didn't accidentally zoom in with your browser, just reset the zoom level. If the zoom level isn't causing it then let me know what browser and OS are you using so I can test it. Thx

AndonIgnatov:
regarding that white space on the right when in mobile view, that's created by the "Leader Board" table below my component since it's a table. My component is fully responsive. You can test it without the Leader Board if you edit the Dashboard.jsx and comment out `<Leaderboard />`

PR #25, opened 2020-09-08
Wbs2.1 query tasks by userid

jerrydevs:
Add getTasksByUserId to taskController.

Can query all tasks assigned to a particular user.


chrisweilacker (approved):
Looks good.

PR #32, opened 2021-01-10
feat: Added the collaborationPreference column in user profile

getnithesh:
A new column has been added to the user profile for capturing the collaboration preference of a user.


daoquangvy (approved):
Everything seems to work fine. I will approve this.

PR #39, opened 2021-02-05
Fixed Email Sending to Allow More Than 16 emails

chrisweilacker:
Fixed Email Sending to Allow More Than 16 emails through the use of a timer on the infringement emails.


chrisweilacker:
Also undid two of the warning fixes that should allow for the page to have a faster operation.

getnithesh (approved):

PR #40, opened 2021-02-06
Wbs4.2.1/forgot password email update

getnithesh:


daoquangvy (approved):
1. Everything seems to work normally besides sending an actual email to the user email. However, it does not cover under the scope of this branch at this moment. So I will approve this.

PR #43, opened 2021-02-15
update task

vnseattle:
Update Task
Fix bugs


daoquangvy (approved):
The 404 Not Found bug is still exits. However, I will approve this and Vignesh will take care of this in the future.

PR #44, opened 2021-02-27
Leaderboard org data route

chrisweilacker:
Completed the Org Data Route which gets the entire Organization leaderboard data so users can see the entire leaderboard totals rather than just those of their team.


daoquangvy (approved):
Similar to https://github.com/OneCommunityGlobal/HighestGoodNetworkApp/pull/127

Everything is working, I don't see any error, bugs.

I will approve this one.

PR #47, opened 2021-03-09
Bug Fix: Fixed bug where if a user is in no teams they aren't seen in their leaderboard

chrisweilacker:
Added functionality to get the User's information sent in cases where a user is in no teams.


daoquangvy (approved):
Everything seems to work the way it supposed to be, I don't see any bugs or errors. I will approve this.

- I have attached the picture below as proof that no team user will also show up in Leaderboard ( The top is the current branch, the bottom is this branch) :heavy_check_mark:

image

PR #70, opened 2021-08-11
Weekly summary report feature parity

CameronWhiteCS:
- The list of active volunteers' emails now appears at the bottom of emailed weekly summary reports as requested

- The information contained in the PDF and in-browser versions of weekly summary reports is now included in the email variant as well 

- Currently, the server doesn't provide the client with access to the emails associated with weekly summaries preventing the report generation feature from working properly. This is now fixed.

- The weekly committed hours associated with each weekly summary is now sent to the client when requesting weekly summary data


chrisweilacker (approved):
lgtm

PR #76, opened 2021-08-20
Fixed a bug where the leaderboard displays incorrect values

CameronWhiteCS:
Fixes issue: 

> Jae: Dashboard → Leaderboard → Looking at the HGN Totals (see below)
This is not totalling the amount of committed time (1925.8 in the pic below) correctly. This total should be a sum of all active members’ weekly “tangible time” committed. 
Jae 8/20: I found part of the problem here, maybe all of it. Every time I add intangible time it increases the logged time (190.57 in the pic below) by the same amount and the total time (1925.8 in the pic below) by 10x whatever the intangible time added was - intangible time shouldn’t count at all into either of these numbers. 
E.g. I logged 2 hrs intangible time and the numbers on the right went from 254.33 of 2473.2 to 256.33 of 2493.2 (Cameron: PR #224)


chrisweilacker (approved):
lgtm

PR #82, opened 2021-08-28
Fixed issues with time entry edit emails

CameronWhiteCS:
Fixes issue:
> Bug1) The popup says 6 or more times gives blue squares (see pic below) and it actually gives them after 5 or more times. See pic below that shows 7 edits and 3 blue squares >> Please update the text to say "5 or more".  >> I also did not get an email notifying me that a blue square was issued for this. I'm guessing they wouldn't have either.


chrisweilacker (approved):
lgtm

PR #83, opened 2021-08-30
Leaderboard fixes 2

CameronWhiteCS:
Fixes issue:

> Cameron: The leaderboard still needs some adjusting. PR 243
“HGN Totals: ___ Members” not accurate - shows team size and not size of org
This one looks good for how it shows the whole team, but my test account in the Dev doesn’t show my hours in the leaderboard now. See screenshot below. 

The function ```getUserLaborData``` in ```helpers/dashboardhelper.js``` previously used some fairly complicated aggregation logic. Due to its complexity, it was relatively error prone and hard to debug. This PR replaces that aggregation with two queries and a for loop and fixes the above issue.

### DIFF
--- a/src/helpers/dashboardhelper.js
+++ b/src/helpers/dashboardhelper.js
@@ -1,26 +1,26 @@
-const moment = require("moment-timezone");
-const mongoose = require("mongoose");
-const userProfile = require("../models/userProfile");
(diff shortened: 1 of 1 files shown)


one-community (approved):
Approved while on screen share with Cameron

Issue #87, opened 2021-08-31
Cron jobs not guaranteed to completed in specific order


CameronWhiteCS:
Despite needing to execute in a specific order, the cron jobs for HGNRest are not guaranteed to do so.

Jobs are scheduled a few minutes apart under the assumption that job A will be finished by the time job B begins. 

Solution: refactor the existing cron jobs into a single job that uses async / await to guarantee jobs finish in the desired order, possibly even more quickly than they currently do.

one-community:
I think the reason it is set up to process in a specific order like that is because Google will only allow sending so many emails at once. So having them all sent at the same time will mean Gmail gets blocked or doesn’t send them at all if there are a lot to send at once. Chris said we should switch away from gmail at some time because of this and security reasons. Now may be the time. 

What do you think?

CameronWhiteCS:
That's not what I was referring to. The order of the events is fine, but the mechanism that the ordering of the events is based off of isn't guaranteed to work and is more likely to fail as the app gets more users.

one-community:
Ok, thank you for the clarification. Just wanted to make sure we’re not changing the way the blue square emails are spread out because that’ll create very bad bug if we do.

CameronWhiteCS:
If Job A needs to run before B, and B before C, the way the system is currently designed, it doesn't check if A has finished executing before running B -- it just blindly runs.

one-community:
Got it. That’d definitely cause the problem we have. Especially since I don’t think the “unpause” people step is scheduled to happen after the send blue squares step… I think it is currently set up as an independent action that runs right at midnight for whatever day they are paused to, and, in the case of Saturday nights, the blue squares all then run 5 minutes later.

PR #90, opened 2021-09-02
Fixed ordering of Cron Jobs

CameronWhiteCS:
Blue squares are now issued for failing to meet the number of weekly committed hours BEFORE inactive users are marked as active. 

Solves issue:
> Jae: Person’s Profile → Basic Information → Pause → Choose Date
I paused someone for the first time last week and had them set to restart today (Sunday). They still received a blue square. My guess as to why is because the blue squares go out at 12:05 AM and the pause turned off right at 12 AM. We don’t want to change the blue squares going out at 12:05 though (ask Chris for why), so can we make the “pause” end at 12:15 AM PDT on whatever day a person is paused to?

### DIFF
--- a/src/cronjobs/userProfileJobs.js
+++ b/src/cronjobs/userProfileJobs.js
@@ -3,13 +3,6 @@ import { CronJob } from 'cron';
 const userhelper = require('../helpers/userhelper')();
 
 const userProfileScheduledJobs = function () {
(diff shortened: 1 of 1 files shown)


calvertwong (approved):
Approved the temp fix, the better solution is to refactor the code as suggested in this thread
https://github.com/OneCommunityGlobal/HGNRest/issues/87

PR #95, opened 2021-09-07
Fix/insert volunteer end date

calvertwong:
Added endDate field to store the date that a volunteer is being paused by an admin. Added a date validator to add the endDate to Mongo if valid and remove the endDate field from Mongo if the date input is empty which happens when the admin removes the endDate from the form.

image

### DIFF
--- a/src/controllers/userProfileController.js
+++ b/src/controllers/userProfileController.js
@@ -1,3 +1,5 @@
+import yearMonthDayDateValidator from '../utilities/yearMonthDayDateValidator';
+
 const mongoose = require('mongoose');
 const bcrypt = require('bcryptjs');
 const moment = require('moment-timezone');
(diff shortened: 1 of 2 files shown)


src/controllers/userProfileController.js
```diff
@@ -228,6 +230,11 @@ const userProfileController = function (UserProfile) {
         record.timeZone = req.body.timeZone;
         record.hoursByCategory = req.body.hoursByCategory;
         record.createdDate = moment(req.body.createdDate).toDate();
+        if (yearMonthDayDateValidator(req.body.endDate)) {
```
calvertwong, line 233: Date input format must be YYYY-MM-DD to pass the check

src/controllers/userProfileController.js
```diff
@@ -228,6 +230,11 @@ const userProfileController = function (UserProfile) {
...
+          record.endDate = moment(req.body.endDate).toDate();
+        } else {
+          record.set('endDate', undefined, { strict: false });
```
calvertwong, line 236: As for now, remove the endDate field if the data input is not valid. Currently the UI does not have the date validator and allows any invalid date input to submit but the UI will later get an error dialog.

CameronWhiteCS (approved):

PR #109, opened 2021-09-14
Fixed weekly summary reports

CameronWhiteCS:
Fixes issue: 
> The weekly summaries in the app (Reports → Weekly Summaries Report) are now showing total time instead of just tangible too. Below is my example for last week. It should say 32.38 / 40 since I only did 32.38 hours of tangible time.


chrisweilacker (approved):
lgtm

PR #130, opened 2022-02-28
Send Timezone API key based on user class

pranav1416:
This PR creates a new route/controller in the app to integrate OpenCage API to fetch the timezone based on the input location provided by the user. 
A new route and controller is added to send protected (currently free) API key to the frontend application. 
Note - currently we are using a freemium account on OpenCageData and have limited number of requests available to us in one single day (2500 API requests per day) which are quite sufficient since this functionality is currently only used in creation of new users form.

### DIFF
--- a/src/controllers/timeZoneAPIController.js
+++ b/src/controllers/timeZoneAPIController.js
@@ -1,0 +1,23 @@
+const timeZoneAPIController = function () {
+    const getTimeZoneAPIKey = (req,res) => {
+        const requestorRole = req.body.requestor.role;
+        const PremiumRoles = ['Manager', 'Administrator', 'Core Team'];
+        const isPremiumRequestor = !!((PremiumRoles.includes(requestorRole)));
+        const premiumKey = process.env.TIMEZONE_PREMIUM_KEY;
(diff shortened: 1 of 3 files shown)


src/controllers/timeZoneAPIController.js
```diff
@@ -0,0 +1,23 @@
+const timeZoneAPIController = function () {
+    const getTimeZoneAPIKey = (req,res) => {
+        const requestorRole = req.body.requestor.role;
+        const PremiumRoles = ['Manager', 'Administrator', 'Core Team'];
+        const isPremiumRequestor = !!((PremiumRoles.includes(requestorRole)));
+        const premiumKey = process.env.TIMEZONE_PREMIUM_KEY;
+        const commonKey = process.env.TIMEZONE_COMMON_KEY;
```
pranav1416, line 7: I need to configure deployment service to add these two variables. Currently I will be adding same key for both of these use cases, but this feature can be used on each user profile in future in which case we can have premium key access to selected user classes and add the premium key to the deployment configurations when we are ready for it. 
@chrisweilacker I might need your help to add those configurations and also please review this PR once you get a chance, thanks!

getnithesh (approved):

PR #137, opened 2022-04-15
fix issue where privacy settings wasn't being cached correctly

nchen0:
- Noticed privacySettings and hoursByCategory both weren't caching correctly, according to https://github.com/node-cache/node-cache/issues/77, I think it's because node-cache doesn't pick up nested objects correcly, which those two fields are.

### DIFF
--- a/src/controllers/userProfileController.js
+++ b/src/controllers/userProfileController.js
@@ -354,7 +354,7 @@ const userProfileController = function (UserProfile) {
   const getUserById = function (req, res) {
     const userid = req.params.userId;
     if (cache.getCache(`user-${userid}`)) {
-      const getData = cache.getCache(`user-${userid}`);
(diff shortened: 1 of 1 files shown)


chrisweilacker (dismissed):
lgtm

chrisweilacker:
> * Noticed privacySettings and hoursByCategory both weren't caching correctly, according to [[QUESTION] Nested key:value store?  node-cache/node-cache#77](https://github.com/node-cache/node-cache/issues/77), I think it's because node-cache doesn't pick up nested objects correcly, which those two fields are.

Maybe should look at the other spots where it is being cached only think there are a couple to change those.

nchen0:
@chrisweilacker Good call out, added it to the other place where I see cache being set.

PR #138, opened 2022-04-16
make privacy settings default to true & fix location bug

nchen0:
- Add location to user profile schema to fix bug in frontend where location doesn't persist


chrisweilacker (approved):
lgtm, assuming it works when tested

PR #140, opened 2022-04-20
Fix email subject spelling error

aChrisChen:
Update mis-spelling subject in email template.

### DIFF
--- a/src/helpers/userhelper.js
+++ b/src/helpers/userhelper.js
@@ -338,7 +338,7 @@ const userhelper = function () {
 
           emailSender(
             status.email,
-            'New Infringment Assigned',
+            'New Infringement Assigned',
             getInfringmentEmailBody(
               status.firstName,
               status.lastName,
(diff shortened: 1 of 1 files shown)


EvianTan:
I'm not sure the scope of this PR. I do find many misspelling "Infringment" both in HGNRest and HGNApp , do we want to change all of them in this PR?

one-community:
**PR for: Just realized there is a spelling error in the subject line of the automated blue square emails we send out. Infringment should be Infringement.**

Let's get this small change done and then please add/log a separate bug for changing all the misspellings because missing one in the code could break something pretty important and I don't want to risk that this week when I'm traveling. Testing these is also difficult to do.

aChrisChen:
there are lots of misspelling of the “Infringment” as the variable name in the current code base. I would suggest we create a separate task to refractor all of them instead of changing all the strings in this PR.

EvianTan:
> I also noticed that there are lots of misspelling of the “Infringment” as the variable name in the current code base. I would suggest we create a separate task to refractor all of them instead of changing all the strings in this PR to make sure we are not breaking any of functions.

Sounds cool. Please add a separate bug in the list as refactoring

EvianTan (approved):

PR #144, opened 2022-05-06
remove duplicate phone limitation and update backend cache when adding and deleting new user

stevenwang213:
For the ones who review this:
This PR fixes two bugs.
- remove the duplicate number check when creating new user.
Instead of removing the duplicate number check code, I create a new variable `duplicatePhoneNumberCheck` to determine if we need the duplicate phone check. If we need the function in the future, just change the value to true.

- update the backend cache when creating and deleting a user. 
After creating a new user or deleting a user the result will immediately show on the UserManagement table.

### DIFF
--- a/src/controllers/userProfileController.js
+++ b/src/controllers/userProfileController.js
@@ -190,8 +190,12 @@ const userProfileController = function (UserProfile) {
         res.status(200).send({
           _id: up._id,
         });
+        //remove backend cache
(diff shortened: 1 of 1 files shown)


EvianTan:
Tested creating new user and assigning a user to team, both show up immediately. Good Job!

EvianTan (approved):

PR #158, opened 2022-05-25
Improve cache behavior when creating/deleting/updating user data

stevenwang213:
This PR focuses on backend cache improvement.
This PR modifies the cache based on different user management operations e.g. creating, deleting, and updating user profile, rather than removing the whole user cache. 
By doing this, we can save the loading time of the app and eliminate the need to access the database whenever we just make a slight change to one user profile.

### DIFF
--- a/src/controllers/userProfileController.js
+++ b/src/controllers/userProfileController.js
@@ -280,7 +280,7 @@ const userProfileController = function (UserProfile) {
         record.createdDate = moment(req.body.createdDate).toDate();
         if (yearMonthDayDateValidator(req.body.endDate)) {
           record.endDate = moment(req.body.endDate).toDate();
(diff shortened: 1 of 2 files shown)


migfernandes01:
While testing this, I couldn't add and update a user. I did `git pull` to make sure I was using the most recent changes. Delete works though and it looks good!
Here are the errors:
error creating user
error updating user

stevenwang213 (commented):
I tried "add" and "update" again, but I didn't get any error at local. Did you get any error messages in the terminal?

migfernandes01:
I ran `npm run serve` this time and it worked all well!

migfernandes01 (approved):
Passed all tests

PR #160, opened 2022-06-08
add hoursLogged field on task schema

EvianTan:
This PR is the backend change for "implement the three clock icons on management-dashboard".

**More details:**
The green and red clocks both need the logged hours data of each task, so created `hoursLogged` field in task's schema.


thienphuvn95 (approved):
Approved

PR #165, opened 2022-06-24
update comment

EvianTan:
This PR is the backend changes for showing taskname of each timeEntry on timelog.

This change solves when calling `getTimeEntriesForSpecifiedPeriod()`, the task-based timeEntry data cannot get `taskId`. After change, the `taskId` can also be populated into the `projectId` field, which is same as project-based timeEntry.


kimsj23 (approved):
Looks good. Approved

kimsj23 (approved):
approved

kimsj23 (approved):
approved

PR #166, opened 2022-06-28
Task notification on adding and editing a task with modal differences

esk47:
Dependent on [PR 463 on HighestGoodNetworkApp](https://github.com/OneCommunityGlobal/HighestGoodNetworkApp/pull/463)

See that PR to test.


EvianTan:
I just think about a corner case. When editing a task, if someone is unassigned from this task, he or she should also be notified that. But in our current management-dashboard it cannot have that kind of notification for unassigning. So I think put this corner case in our management bug list is good for now.

EvianTan (approved):

PR #173, opened 2022-07-22
update the timeEntry controller for showing hoursLogged

EvianTan:
This PR is completing the "log time to assigned tasks for current user" feature.
I updated those three functions in `timeEntryController.js`:
- editTimeEntry()
- postTimeEntry()
- deleteTimeEntry()

Steps to test:
1. Make couple new tasks to make sure them have `hoursLogged` field (most of the old tasks don't have this field so they won't work there)
2. Log tangible/intangible time slots to tasks (**only** tangible time will be counted into `hoursLogged` field of task)
3. Play with post/delete/edit time slots. **Don't forget hard refreshing page to get actions effected. I'll create a bug/improvement for this.**

See below image for result:
image

### DIFF
--- a/src/controllers/timeEntryController.js
+++ b/src/controllers/timeEntryController.js
@@ -89,6 +89,7 @@ const timeEntrycontroller = function (TimeEntry) {
 
       const initialSeconds = timeEntry.totalSeconds;
       const initialProjectId = timeEntry.projectId;
+      const initialIsTangible = timeEntry.isTangible;
 
       timeEntry.notes = req.body.notes;
       timeEntry.totalSeconds = totalSeconds;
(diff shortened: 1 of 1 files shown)


esk47:
Taking a look

esk47:
There is a bug with editing time entries.

If you post an intangible time entry, then convert it to tangible, the hours logged doesn't increase.
If you then delete that same time entry, the hours logged decreases.

This can lead to negative hours being displayed for a task:
image

EvianTan:
> There is a bug with editing time entries.
> 
> If you post an intangible time entry, then convert it to tangible, the hours logged doesn't increase. If you then delete that same time entry, the hours logged decreases.
> 
> This can lead to negative hours being displayed for a task: image

Good catch! I'll rethink the logic to revert negative hours showing.

esk47:
I think the bug is more on the side of editing a time entry from intangible to tangible not tracking on hours logged.

EvianTan:
> I think the bug is more on the side of editing a time entry from intangible to tangible not tracking on hours logged.

Yup. If it's need too much work, I'll file it as a bug 😄 .

esk47 (approved):
Rework fixes everything!

I'm a little confused by the conditionals. Also I think initialProjectId and req.body.projectId are always the same so I think the first conditional could be more concise.

Approving this, but let me know you think.

EvianTan:
> Rework fixes everything!
> 
> I'm a little confused by the conditionals. Also I think initialProjectId and req.body.projectId are always the same so I think the first conditional could be more concise.
> 
> Approving this, but let me know you think.

Oh, I thought this one was confused too, let me explain that. All the line numbers I mention below are the latest code ones.
1️⃣ we got the before/initial `timeEntry` and stored on line 71, and got its id as `initialProjectId` on line 91
2️⃣ `timeEntry.projectId` is overwritten by `req.body.projectId`(which is the after/edit timeEntry) on line 98

In the first condition (line 103-127), `initialProjectId` is the before/initial timeEntry id, `req.body.projectId` is the after/edited timeEntry id. They're different and both needed.

esk47:
Sounds good!

PR #176, opened 2022-08-13
fix: saving updated timezone does not update in database due to user …

rutvij2811:
Issue: Updating the profile with the new timezone is not reflected in the UI or the database.

Root cause: The method/endpoint "putUserProfile" checks if the role has the permission "putUserProfileImportantInfo" before deciding if he/she is allowed to update the timezone. As for most of the cases, the user role will be "Volunteer", which does not have the permission "putUserProfileImportantInfo". Hence the user won't be able to save the new timeZone.

Resolution: Either give the "Volunteer" role the required permission or shift the timezone update to the generic field update section before this role check section. Option 1  is not feasible as the "putUserProfileImportantInfo" many admin related permissions.

Behaviour After Fix:
saveTImeZoneFix

Note: This is a partial fix for bug 4 in the bug doc.  The other fix is to be taken care of in the front-end as no timezone is passed during the user creation as a result of which the default "America/Los_Angeles" is stored for all the new users.

### DIFF
--- a/src/controllers/userProfileController.js
+++ b/src/controllers/userProfileController.js
@@ -246,6 +246,7 @@ const userProfileController = function (UserProfile) {
       record.weeklySummaries = req.body.weeklySummaries;
       record.weeklySummariesCount = req.body.weeklySummariesCount;
       record.mediaUrl = req.body.mediaUrl;
+      record.timeZone = req.body.timeZone;
 
       // find userData in cache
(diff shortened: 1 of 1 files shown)


nickychen0 (approved):

EvianTan:
Nice job! The description above is pretty clear, thank you Rutvij!

EvianTan (approved):

PR #179, opened 2022-08-16
[Feature] Timer service integration work for real-time syncing

andrewpaulino:
### Timer Service Backend

Hello, this PR introduces the foundational work to make the backend run smoothly for the new timer service. It uses a mix of MongoDB, WebSockets, and a key-value database called Redis.

From an architectural standpoint, the client will be automatically connected to the websocket server -- which will then add an active connection in the Redis service. We use Redis for any active connections / active timers, and use MongoDB for long storage of timer data.

**Some features encapsulated now**

* Real-time sync between all timers based on user
* Automatic detection if user's timer went to sleep
* Heartbeat websocket system
* Fault-tolerant clean up system, we will save all active timers if server goes down
* Scalable websocket server / express server. We can add as many servers, since our data lives in redis.
* Pub-Sub system to relay updates to all clients subscribing to timer.

This feature should be backwards compatability, and the leverage we will have will be on the front-end.

Untitled-2022-08-27-0957


EvianTan:
That's an awesome feature! 👍 

For the initial review, please help me understand, why we need `timer` and `oldTimer`. I guess `oldTimer` is for MongoDB to save a stopped timer object data which `timer` is for Redis to store a temporary object data. Does my guess sounds right?

andrewpaulino:
> That's an awesome feature! 👍
> 
> For the initial review, please help me understand, why we need `timer` and `oldTimer`. I guess `oldTimer` is for MongoDB to save a stopped timer object data while `timer` is for Redis to store a temporary timer object data like pausing by user or sleep. Does my guess sounds right?

Well yes and no? oldTimer is being used to support the older timer functionality. With our new schema it would be hard to support the current timer in the production. This way we can deprecate the timer as we go along with the feature.

EvianTan:
> > That's an awesome feature! 👍
> > For the initial review, please help me understand, why we need `timer` and `oldTimer`. I guess `oldTimer` is for MongoDB to save a stopped timer object data while `timer` is for Redis to store a temporary timer object data like pausing by user or sleep. Does my guess sounds right?
> 
> Well yes and no? oldTimer is being used to support the older timer functionality. With our new schema it would be hard to support the current timer in the production. This way we can deprecate the timer as we go along with the feature.

Cool, got that 😄

Programmingrelated1999:
Based on my understand from reading previous pages and diagrams, the current timer has a different schema with MongoDB. So we use web socket to communicate and store data on REDIS. When the timer ends, the data from current timer is transformed to be store in MongoDB? Please let me know if I am missing something here.

src/websockets/TimerService/utility.js
```diff
@@ -0,0 +1,63 @@
...
+export const calculateTotalSecondsBasedOnStartingTime = ({ currentTimeInSeconds, startedAtInSeconds, seconds }) => currentTimeInSeconds
+          - startedAtInSeconds
+  + seconds;
+
+export const getTimerFromRedis = async ({
+  userId,
+  redisClient,
+} = {}) => {
+  try {
+    return JSON.parse(await redisClient.get(userId) ?? '{}');
```
esk47, line 14: If userId == null, do we want to do get with an empty object?

It seems like throwing an error is good here.

src/websockets/api.js
```diff
@@ -0,0 +1,218 @@
...
+
+export const redisClients = {
+  main: createClient({
+    url: process.env.REDIS_APP_URL,
+  }),
+  publisher: createClient({
+    url: process.env.REDIS_APP_URL,
+  }),
+  subscriber: createClient({
+    url: process.env.REDIS_APP_URL,
```
esk47, line 19: Should main, publisher, and subscriber have the same url?

esk47 (commented):
Everything looks great.

src/websockets/api.js
andrewpaulino: Yes, this is all part of redis

src/websockets/TimerService/utility.js
andrewpaulino: I mean from a front-end perspective nothing would change.

src/controllers/timerController.js
```diff
@@ -1,4 +1,62 @@
...
+      const timerObject = await Timer.findOne({ userId }).exec();
+      if (!timerObject) {
+          const newRecord = {
+            userId,
+            totalSeconds: 0,
+            isRunning: false,
+            isApplicationPaused: false,
+            isUserPaused: false,
+          };
+          const newTimer = await Timer.create(newRecord).exec();
```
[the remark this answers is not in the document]
andrewpaulino: Good question, basically this is if there is no timer instance from the user in the database. Usually this could be since the user is new, there has been no timer created yet. 

`setTimerToDatabase` Updates a record, which assumes the timer already exist.

src/websockets/api.js
```diff
@@ -0,0 +1,218 @@
...
+  }),
+};
+
+export const intents = {
```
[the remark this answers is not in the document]
andrewpaulino: Nope, good catch. Will remove export.

src/websockets/TimerService/utility.js
```diff
@@ -0,0 +1,63 @@
+import { TIMER_UPDATES_CHANNEL } from 'websockets/api';
+
+const logger = require('../../startup/logger');
+
+export const calculateTotalSecondsBasedOnStartingTime = ({ currentTimeInSeconds, startedAtInSeconds, seconds }) => currentTimeInSeconds
+          - startedAtInSeconds
+  + seconds;
```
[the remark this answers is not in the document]
andrewpaulino: Yup, I think the linter messed this up.

src/models/timer.js
```diff
@@ -4,10 +4,10 @@ const { Schema } = mongoose;
 
 const timerSchema = new Schema({
   userId: { type: Schema.Types.ObjectId, required: true, ref: 'userProfile' },
```
Hani6148, line 6: Try to fix timer schema name.

Hani6148 (approved):

PR #187, opened 2022-09-09
DeActivateUser function initiation.

kseniaxu0704:
DeActivateUser function :

Aim of DeActivateUser function: after the date reaching out the final day, the user's status will change from active to inactive. 

How to achieve:

file userhelper

1) Find the user who meets the condition that needs to be deactivated using mongoose - useprofile. find();(three conditions: currently isActive, have already set endDate, the current date is the same or later than the endDate);

2) After finding the user then update the user information and change the user' isActive from true to false; (use findOneAndUpdate())

file userProfileJobs:

3) This function will execute every day one minutes past midnight (use cronjob )

In this PR,  I add a test function in the comment, if you want to test the effect of this function, you can delete the /**/ to test it. The test function is totally the same as the deActiveUser function, except the current date which changes to the current date plus five days, so if you select a final day within five days, it could be a test. 

The test function explanation  as below:

In the real situation, if today is reaching the final day, the function will set the user’s status from active to inactive. But in the test situation, since we want to see the effect immediately, I write a function like this moment().add(5, days), which means if today plus five days is later than the final day, the function will change from active to inactive. Therefore we can test future days. For example, 
End Date : Sep 10
Today: Sep 9,
Test day :  Today(Sep 9) + 5 days =  Sep 14,
In the real situation: if today is later than endDate, the function will be called.
In the test situation: Test day is later than endDate. Therefore function will be called. 

What is more, the deactivated user function should be executed every day one minute past midnight, but in the test situation, since we want to see the effect immediately, I write a function, then it is possible to execute every minute we want.

### DIFF
--- a/src/cronjobs/userProfileJobs.js
+++ b/src/cronjobs/userProfileJobs.js
@@ -15,6 +15,7 @@ const userProfileJobs = () => {
         await userhelper.awardNewBadges();
       }
       await userhelper.reActivateUser();
+      await userhelper.deActivateUser();
     },
     null,
     false,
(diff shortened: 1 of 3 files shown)


Programmingrelated1999:
I have the DeActivateUser function working on my computer. I was able to successfully deactive user as shown below. 
PR_review3_Yan_backend
I will go ahead and approve the PR request.

Programmingrelated1999 (approved):

PR #189, opened 2022-09-15
fix bug for deActivateUser fucntion

kseniaxu0704:
Fix Bug for deactivate user function:

In the past , If the final day is 09/15, therefore user will deactivate on 09/15;
Now, if the final day is 09/15, therefore user will deactivate on 09/16;
use moment().isAfter(moment(endDate);

In addition, i delete all the test function in comment;

### DIFF
--- a/src/cronjobs/userProfileJobs.js
+++ b/src/cronjobs/userProfileJobs.js
@@ -23,17 +23,5 @@ const userProfileJobs = () => {
   );
 
   allUserProfileJobs.start();
-/* For Test
-  const test1 = new CronJob(
-    '42 * * * *', // At minute 42. You can change the minute whenever your want to test 
-    async () => {
-      await userhelper.test();
-    },
-    null,
-    false,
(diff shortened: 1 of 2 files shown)


Programmingrelated1999:
I was able to get it work on my computer. 

The terminal prints findAndModify isActive as false for tomorrow (sept 17) and true for other days after sept 17. I notice that the test functions are commented out but may I know which function is printing the logging info as shown below. 

PR_yan1

Programmingrelated1999 (approved):

kseniaxu0704:
Hi Kaung, 
Two things may cause this problem:
Firstly, every time we change the REST, if we want to test it, we need first to run npm install, then npm run build and then npm start; just a single npm start does not work.
image
Secondly, the branch may be different to cause this issue. 
If it's neither, please let me know, and I'll fix it.

Programmingrelated1999:
I got it working after npm run build. Thanks for the help!

PR #203, opened 2022-10-11
[Backend] Revert timer system

andrewpaulino:
As per the discussion with Jae, we would like to revert some of the timer work to make sure no bugs are in the system till we can get the work in a stable position.

I reverted back to the old way, and kept the real-time timer code but commented out the start of the websocket server so nothing will run. If someone wants to revisit it it is there and here is the original PR for it. https://github.com/OneCommunityGlobal/HGNRest/pull/179/files#


Red4lpha (approved):
Looks good, timer works, build successfully.

PR #207, opened 2022-10-14
Fix projects pie chart

veratim1996:
Passing from the backend ids of the projects to which users tasks belong, in order to display the pie chart of the projects correctly.


one-community (approved):

EvianTan:
Code looks great and test with no weird things. The only concern is I cannot find or reproduce the problem in description.

EvianTan (approved):

PR #219, opened 2022-11-05
Permissions dashboard - new routes

guiwustro:
Creation of new routes:
- GET /roles - should return all roles and permissions back-end/front-end of each;
- POST /roles - should be able to create a new role just if is Owner;
- PATCH /roles/:id - should be able to updatte a role if is Owner
Also, add new array in the token, the permissions, that contains the front-end permissions and the back-end permissions of an user. This array is add in the User model.
These changes are necessary to build a permissions Management of the users. The old management was in an Array with all permissions mocked.


KaixiangGU (approved):
I tested with an owner account, everything works well.

PR #225, opened 2022-11-19
Fix/owner permissions

guiwustro:
Changed the "for each" for "for loop" using async functions .
Reassigning the owner role to Jae .

### DIFF
--- a/src/utilities/createInitialPermissions.js
+++ b/src/utilities/createInitialPermissions.js
@@ -283,7 +283,7 @@ const permissionsRoles = [
 
 
 const createInitialPermissionsFront = async () => {
-  const userEmail = { email: 'jae@onecommunityglobal.orgg' };
+  const userEmail = { email: 'jae@onecommunityglobal.org' };
   const update = { role: 'Owner' };
   await User.findOneAndUpdate(userEmail, update);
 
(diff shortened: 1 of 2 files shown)


src/utilities/createInitialPermissions.js
```diff
@@ -279,44 +280,64 @@ const permissionsRoles = [
...
         }
-        if (!IsAllBackPermissionsOwnerUpdated) {
-          await Role.findById(roleId, (_, record) => {
-            record.permissionsBackEnd = permissionsBackEnd;
-            record.save();
-          });
+        if (!IsAllFrontPermissionsOwnerUpdated) {
+        updatedOwnerPermissions.permissionsBackEnd = permissionsBackEnd;
         }
       }
```
KaixiangGU, line 317: These two conditional statements are the same, is the second one supposed to be "!IsAllBackPermissionsOwnerUpdated"?

guiwustro: Yess, you are right, already changed to the correct way. Thank you

KaixiangGU: Everything else looks good to me! Going to approve it.

KaixiangGU (approved):

PR #248, opened 2022-12-10
multi bug fix and add new emailreminder function for RE/DE-Activate user

kseniaxu0704:
This PR focus on two function - reActivateUser&deActivateUser function
For reActivateUser :
1) add email reminder - when reaching out reactivate date, admin will receive a email and formate like that 
image
2) delete the end date when reaching out the reactivate user using $unset MongoDB
For deActivateUser:
1) fix email fomate 
2) in the past, it is when reaching out the final day, the user change from active to inactive ;
but now it is pass the final day(end date), then user change from active to inactive.

Note: if you want to verify in your local, and is not familiar with Oauth2.0, please feel free to slack me. Thanks.

### DIFF
--- a/src/helpers/userHelper.js
+++ b/src/helpers/userHelper.js
@@ -1,6 +1,5 @@
 const mongoose = require('mongoose');
 const moment = require('moment-timezone');
-const _ = require('lodash');
 const userProfile = require('../models/userProfile');
 const timeEntries = require('../models/timeentry');
 const badge = require('../models/badge');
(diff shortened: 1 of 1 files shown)


src/helpers/userHelper.js
```diff
@@ -984,7 +1007,7 @@ const userHelper = function () {
         const user = users[i];
         const { endDate } = user;
         endDate.setHours(endDate.getHours() + 7);
-        if (moment().isAfter(moment(endDate))) {
+        if (moment().isAfter(moment(endDate).add(1, 'days'))) {
```
KaixiangGU, line 1010: Hi Yan, everything else looks good. Just feel a little confused about these lines.

For now, users will be deactivated after the endDate right? For example, a user's end date is Dec-10 and this account will turn inactive on Dec-11. 
From my understanding, `isAfter` has already checked if today's date is after the end date. If you add 1 more day on top of the end date here, the above example will not be true on Dec-11, it will be true on Dec-12. Because Dec-11 is ****NOT** after** endDate+1day which is also Dec-11.
Correct me if I was wrong!

KaixiangGU: Thanks, Yan, that makes a lot more sense. Good Job 👍

KaixiangGU (approved):
All GOOD

PR #249, opened 2022-12-10
Changed some conditions on the formatting of summaries.

LeeSingChanYau:
Weekly summaries are registering as submitted on the Dashboard (see pic below) even when they haven’t been. This appears to be happening because the submission page is one week off from the correct date. This does not include the additional request. Please test this before merging.

### DIFF
--- a/src/controllers/reportsController.js
+++ b/src/controllers/reportsController.js
@@ -1,4 +1,4 @@
-const reporthelper = require('../helpers/reportHelper')();
+const reporthelper = require('../helpers/reporthelper')();
 const hasPermission = require('../utilities/permissions');
 
 const reportsController = function () {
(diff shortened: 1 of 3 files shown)


src/helpers/reporthelper.js
```diff
@@ -169,7 +169,7 @@ const reporthelper = function () {
             wSummaries[1] = null;
             wSummaries[2] = { ...wS[0] };
           }
-        } else { // When two entries.
+        } else if (wS.length === 2) { // When two entries.
```
RJBornToWin, line 172: I see that the code change has been based on the previous comment "When two entries". But the actual file's logic works for 2 as well as 3 since the indexing starts from 0. Since there are critical bugs going, i suggest leaving this as original and be merged in later version.

src/controllers/reportsController.js
```diff
@@ -1,4 +1,4 @@
-const reporthelper = require('../helpers/reportHelper')();
+const reporthelper = require('../helpers/reporthelper')();
```
RJBornToWin, line 1: Critical change. Needs to be merged ASAP as it corrects the incorrect file name.

RJBornToWin (commented):
second change in reporthelper.js can be avoided at this point of firefighting the critical bugs in the app. Or it needs thorough testing before merging.

one-community:
Yiyun said: "Just checked the PR 249, I don't think that one has much value. Feel free to close it for now, I'll take over that PR/bug later"

PR #251, opened 2022-12-16
fix the bug for infringments not match the number of  blue square

kseniaxu0704:
High Priority #6.: Blue square counts stated in emails to team members are not accurate now. Everyone who got an email his week had that email saying it was their 1st blue square, regardless of the number.

image
image

infringement is an array, and each element in this array is named infringement
each infringement has three variables: id, date, description,

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4202,6 +4202,11 @@
         "word-wrap": "~1.2.3"
       }
     },
+    "ordinal": {
+      "version": "1.0.3",
+      "resolved": "https://registry.npmjs.org/ordinal/-/ordinal-1.0.3.tgz",
+      "integrity": "sha512-cMddMgb2QElm8G7vdaa02jhUNbTSrhsgAGUz1OokD83uJTwSUn+nKoNoKVVaRa08yF6sgfO7Maou1+bgLd9rdQ=="
+    },
     "os-shim": {
       "version": "0.1.3",
       "resolved": "https://registry.npmjs.org/os-shim/-/os-shim-0.1.3.tgz",
--- a/package.json
+++ b/package.json
@@ -59,6 +59,7 @@
     "node-cache": "^5.1.2",
     "node-datetime": "^2.0.3",
     "nodemailer": "^6.4.16",
+    "ordinal": "^1.0.3",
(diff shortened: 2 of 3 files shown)


EvianTan:
Hey @kseniaxu0704 ,

I just got time to understand this bug and your review here. 

From my perspective, the problem or bug is not caused by the `getInfringementEmailBody` you changed. Basically the `ordinal()` is the function which convert number to ordinal.

The truth is: our previous infringements record stored under the misspelling `infringments` array, however, one or two weeks ago, we fixed all the misspelling, so the on the `userProfile` controller, we newly created a property called `infringements`. So last Sunday, blue squares were added under correct `infringements` array, which was empty when created. So that one was announced as 1st infringement of 5, since previous records are under `infringments` array.
 
 Below is my "Yiyun Manager" userProfile on Dev as an example:
image

kseniaxu0704:
> 
This branch is about two commits;

The first commit is to check the length of the array (infringements) if is equal to the number of the blue squares; I also connect our dev database using MongoDB 
image
Check if the blue square number in the email equals the length of the infringement array.

The second commit is to improve it because I want to add the st, nd,....suffix in the email and import the package ordinal(); This does not mean I think it is the ordinal cause of the bug.

Glad to hear the old infringement moved to a new array of infringements.

one-community:
Something else we did seems to have fixed this... the count on all the blue square emails that went out this morning is correct. Any idea how that happened? @EvianTan

EvianTan:
@one-community Hey Jae, like I said above, I knew this bug won't happen again today, so I said let's hold this PR until this Sunday. Cause I made changes on Beta DB so this bug won't happen again. It's not a code issue at all. The actual reason is couple days ago we moved all logic under the correct spelling infringements, so database recognizes last Sunday's infringement as the 1st, previous infringements were stored under the misspelling one. Last Thursday I modified them manually on database, so now all good.

one-community:
Got it! So should we just close this PR then?

EvianTan:
@one-community From my perspective, YES. This PR didn't have any functional improvement or issue solutions, but just did a bit code quality improvement.

PR #254, opened 2022-12-18
Feat/guilherme fix delete tasks from user profile + Fix permissionsBackend changing

guiwustro:
This PR will be adding the key "CompletedTask", needed to this PR (where is all explained in there):
https://github.com/OneCommunityGlobal/HighestGoodNetworkApp/pull/603

Also, I'm adding here a FIX in the creation/update a new Role, that I found when I was doing the task, It correct a bugs that will be happening in a future, since the permissionsBackEnd are not changeble in the back-end now. Probably It doesnt show up yet because the feature was not used enough, but I'm sure It will correct this mistake.  
It is IMPORTANT, because in the front-end is sending this data, and the back-end is not changing the "permissionsBackend".
image
(component: PermissionsManagement -> NewRolePopUp.jsx)


EvianTan:
Hi @guiwustro ,

Nice job on this feature. The only reason I cannot approve this PR is, I cannot verify the changes of permission part. I would suggest revert the two lines changes since you already have the PR 255 out.
Once you do the above action. I would approve this PR and also the frontend one.

guiwustro:
> Hi @guiwustro ,
> 
> Nice job on this feature. The only reason I cannot approve this PR is, I cannot verify the changes of permission part. I would suggest revert the two lines changes since you already have the PR 255 out. Once you do the above action. I would approve this PR and also the frontend one.

Thank you. Done!

EvianTan (approved):

PR #255, opened 2022-12-21
Fix/permissions back assignment

guiwustro:
It's a FIX in the creation/update a new Role, that I found when I was doing the task, It correct a bugs that will be happening in a future, since the permissionsBackEnd are not changeble in the back-end now. Probably It doesnt show up yet because the feature was not used enough, but I'm sure It will correct this mistake.
It is important, because in the front-end is sending this data, and the back-end is not changing the "permissionsBackend".

image

(component: PermissionsManagement -> NewRolePopUp.jsx)


EvianTan:
Hey @guiwustro ,

I cannot verify the fix now.
I used my "Yiyun Admin" account on DEV to assign the "handle blue square" permission to my "Yiyun Tan" account(image 1). After logging in and logging out as "Yiyun Tan", this account still cannot edit blue squares(image 2). Can you be more specific like the testing steps and the actions/accounts you used to test?
image
image

Thanks!

guiwustro:
You did correctly! This is the way to test! Sorry for didn't  bring a long test. The change didn't work for you, because the permission to assign new blue squares is "Assign Blue Squares"
image

So, if you remade your steps you will see that will work! I think this label is kind ambiguous saying "Handle Blue Squares", but the permissions handle blue squares allowed you to EDIT/DELETE, not CREATE. I tested here and worked well. I'm also including a video at slack channel!

EvianTan:
All changes can be verified! Love them!

EvianTan (approved):

PR #258, opened 2022-12-25
Bugfix

kseniaxu0704:
Bug description: 

1. When clicking + to manually assign blue square after film the content and saving them, the page shows that An error occurred while attempting to save this profile. 
image
and console show that 

1911671999917_ pic
2. Admins and Users aren’t getting emails now when they are manually assigned a blue square.

What I do :
1. checkout front end and find where will show the alert 
image

2. comment every await function in this try, catch function and then find the first await function existing problem.
3. go to that function, and function Axios put existing problem
image
4. go to the backend end and find the user profile controller, putUserProfile and console log every 400 error in this function.

5. find the notifyInfringements function in the putUserProfile existing problem.
image

Error message suggests that there is no lodash package , but use lodash package, import this package, and then will fix it

### DIFF
--- a/src/helpers/userHelper.js
+++ b/src/helpers/userHelper.js
@@ -1,5 +1,6 @@
 const mongoose = require('mongoose');
 const moment = require('moment-timezone');
+const _ = require('lodash');
 const userProfile = require('../models/userProfile');
 const timeEntries = require('../models/timeentry');
 const badge = require('../models/badge');
@@ -71,7 +72,6 @@ const userHelper = function () {
         <p>Life happens and we understand that. That’s why we allow 5 of them before taking action. This action usually includes removal from our team though, so please let your direct supervisor know what happened and do your best to avoid future blue squares if you are getting close to 5 and wish to avoid termination. Each blue square drops off after a year.</p>
         <p>Thank you,<br />
         One Community</p>`;
-
     return text;
   };
 


EvianTan:
Hey @kseniaxu0704 ,

That's a brilliant bug finding path! I love it. 
I just have two questions:
1. Just curious, I was able to reproduce your steps 1-4, but for the step 5, could you tell me how to reproduce that error?
2. Will this PR also resolve the "manually assign blue squares don't send email" problem?

kseniaxu0704:
> Hey @kseniaxu0704 ,
> 
> That's a brilliant bug finding path! I love it. I just have two questions:
> 
> 1. Just curious, I was able to reproduce your steps 1-4, but for the step 5, could you tell me how to reproduce that error?
> 2. Will this PR also resolve the "manually assign blue squares don't send email" problem?

Hi Yiyun,
Thanks for the comment;
1) Go to the putUserProfile function in the controller, and print the error like that: 
image
then npm install,npm run build, npm start and go to the localhost3000 to assign a blue square, the backend will show that error.
2) Yes, it also resolves the "manually assign blue squares don't send email" problem.

EvianTan:
@kseniaxu0704 .

Thank you so much for the explanation! I can verify that annoying error don't trigger up when editing the user profile and save.
I'll ask Jae to test whether "manually assign blue squares don't send email" solved on Main.

EvianTan (approved):

PR #270, opened 2023-01-14
Johny fix bug changing role from user

johny4437:
# Description
Fixes bug number 3  (bug list medium USER PROFILE COMPONENT)
Changing the role from user profile gets saved and updated on profiles page but doesn’t reflect on other pages/privileges.

## Related PRS (if any):
To test this BackEnd PR you need to checkout the (https://github.com/OneCommunityGlobal/HighestGoodNetworkApp/pull/656) frontend PR.
…

## Mainly changes explained:
When an user make  login in the application  generates a token with some informations like: `user_.id`, `role` and `permissions`, but this token is set just one time, then when a user changed the role for example from volunteer to Manager, this was  being saved in the database but the  ´ token´ wasn't being updated with the new role, then wasn't reflecting on the other pages privilegies, then was created a new function in the `userController` called `refreshToken` now when the user updates the role in the frontend , this function updates the token passing the new  ´role´ and sending back to the frontend.
…

## How to test:
check into current branch
do `npm install`  to run this PR locally.

### DIFF
--- a/src/controllers/userProfileController.js
+++ b/src/controllers/userProfileController.js
@@ -3,6 +3,8 @@ const moment = require('moment-timezone');
 const mongoose = require('mongoose');
 const bcrypt = require('bcryptjs');
 
+const moment_ = require('moment');
+const jwt = require('jsonwebtoken');
 const userHelper = require('../helpers/userHelper')();
 const TimeEntry = require('../models/timeentry');
 const logger = require('../startup/logger');
@@ -10,9 +12,7 @@ const Badge = require('../models/badge');
 const yearMonthDayDateValidator = require('../utilities/yearMonthDayDateValidator');
 const cache = require('../utilities/nodeCache')();
 const hasPermission = require('../utilities/permissions');
-const moment_ = require('moment');
 const config = require('../config');
-const jwt = require('jsonwebtoken');
 
 function ValidatePassword(req, res) {
   const { userId } = req.params;
@@ -661,11 +661,11 @@ const userProfileController = function (UserProfile) {
       res.status(400).send(error);
     }
   };
-  const refreshToken = async(req, res) =>{
-    const {JWT_SECRET} = config;
-    const user = await UserProfile.findById( req.params.userId);
+  const refreshToken = async (req, res) => {
+    const { JWT_SECRET } = config;
+    const user = await UserProfile.findById(req.params.userId);
 
-    if(!user){
+    if (!user) {
       res.status(403).send({ message: 'User does not exist' });
       return;
     }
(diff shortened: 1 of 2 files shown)


rauleffting (approved):
Hi @johny4437! I tested your PR and it worked well, as you can see in the video below:

https://user-images.githubusercontent.com/29555732/212938734-5cabf4f1-e928-4b79-80af-60f014c5dd22.mov

The code is pretty simple and clean, I have nothing to ask about it. The frontend [PR #656 ](https://github.com/OneCommunityGlobal/HighestGoodNetworkApp/pull/656#pullrequestreview-1251769807) was approved as well. Thank you!

johny4437:
Hi @rauleffting , thank you!!

RJBornToWin (changes requested):
Functionally correct. 
A minor suggestion: To match with other routing urls in this router, i think making it "/userProfile/refreshToken/:userId" instead of "/refreshToken/:userId" will be consistent. Correct me if you have made this intentional.

one-community:
Following up on this. I merged the frontend but didn't realize this backend was still needed.

RafaelCastilhoCastro:
Hi @johny4437,
I tested the PR's. Token and permissions are being updated. The feature is working as intended.

johny4437:
@RJBornToWin I did this intentionally, it be easy to change but If I change in the backend now, I would also have to change in the frontend too, but the frontend is already merged.

RJBornToWin (approved):
Approving as the front end was already merged.

PrabhjotSembhi (commented):
hey, I was reviewing this. 

I am getting a dependency conflict error while running "npm install" on this branch. 
image


and when I Tried adding "--force " and "--legacy-peer-deps" with install command to test it.

When I make any changes and try to save the user profile. it throws this error.

Error [ERR_HTTP_HEADERS_SENT]: Cannot set headers after they are sent to the client
image



image

johny4437:
@PrabhjotSembhi this PR was reviewed by two person and was already merged

one-community:
I think this PR may have broken something... now when I delete a blue square from someone (Person's Profile --> Delete Blue Square), my dashboard becomes that person. This is the only backend PR that's been updated recently, so I'm guessing this is the one causing this problem. Here's a video: https://www.dropbox.com/s/gos8pf0mbfnt2jk/Deleting%20blue%20square%20changes%20me%20as%20a%20user%20bug.mov?dl=0

RJBornToWin:
I think this might be related to this PR's associated FE PR @ line 293 in file https://github.com/OneCommunityGlobal/HighestGoodNetworkApp/blob/391ddb4b78cd0a9d08b80088408f7db5d473a303/src/components/UserProfile/UserProfile.jsx 
Also @johny4437 we need to refresh token only if there is a "role change" and if that change is of the same user's profile (not when admin sets the role for another user). During review, I didn't realize (I missed) that the the same JS file is used for self profile and other user profile.

PR #322, opened 2023-04-16
Jinchao fix “Current Week” hours at top of Dashboard Page

suigunfjc:
# Description
image
This bug occurs when users with certain roles, such as managers, have their visibility set to invisible and are not able to view their own leaderboard data. This one-line change will fix the bug.

### DIFF
--- a/src/helpers/dashboardhelper.js
+++ b/src/helpers/dashboardhelper.js
@@ -235,6 +235,7 @@ const dashboardhelper = function () {
                 ],
               },
             },
+            { 'persondata.0._id': userid },
             { 'persondata.0.role': 'Volunteer' },
             { 'persondata.0.isVisible': true },
           ],


one-community (approved):

one-community:
This PR broke the app (blocked everyone from logging in) and had to be reverted. Still need the original problem it is causing fixed though.

suigunfjc:
Hi Jae,

After investigating the error, I don't believe this PR is related to the crash. The red alert in the screenshot appears to be caused by an unsuccessful HTTP GET to the userProfile API [here](https://github.com/OneCommunityGlobal/HighestGoodNetworkApp/blob/development/src/actions/weeklySummaries.js) at line 37 `getWeeklySummaries`, which is not related to the changes made in this PR. From the logs in Sentry, I found that there was a MongoDB network error that occurred at the time of the crash, which I believe is the root cause.

Though I'm confident in this PR, I can open a new PR and complete the necessary testing instructions for the team to review if you would like. Let me know how you would like to proceed.

one-community:
Good news! That would explain why it affected the Dev and Main then too... and why the problems didn't coincide perfectly with the merging of this... What threw me off was the pure coincidence that reverting this PR and fixing of this problem happened within minutes of each other. 

Yes, please raise the PR again and I'll get it checked and merged ASAP.

PR #300, opened 2023-03-22
Jinchao make user invisible to the teams they are on

suigunfjc:
Backend update for frontend PR[731](https://github.com/OneCommunityGlobal/HighestGoodNetworkApp/pull/731)
Add new field to userProfile schema
### How to test
`npm run build`
`npm start`
Test with frontend changes.

PR #327, opened 2023-04-18
Jinchao fix “Current Week” hours at top of Dashboard Page

suigunfjc:
# Description
image
same change as #322 since that one was reverted.
## Main change
In the current dev branch, if a manager or mentor set their visibility to invisible which is the default setting (update from #300 and #[731](https://github.com/OneCommunityGlobal/HighestGoodNetworkApp/pull/731)), the leaderboard data they get from HGNRest won't include their own record, which leads to the bug. This PR adds one line in `getLeaderboard` function to let these users always get their own record whatever their visibility is.
## How to test
In the current branch, log in as a manager or mentor. Go to userProfile->Teams->set visibility to 'invisible'->go back to the basic information tab to save the change. In this case, you won't see yourself in the leaderboard and you won't see your logged hour through the summary bar:
image
Now if you go back and change your visibility to 'visible', everything will look good as usual.
Pull this branch then rebuild and restart the HGNRest. Log in as a manager or mentor, and check if the leaderboard and summary bar looks good whatever the visibility is. (Users should be able to see themselves on the leaderboard)

### DIFF
--- a/src/helpers/dashboardhelper.js
+++ b/src/helpers/dashboardhelper.js
@@ -235,6 +235,7 @@ const dashboardhelper = function () {
                 ],
               },
             },
+            { 'persondata.0._id': userid },
             { 'persondata.0.role': 'Volunteer' },
             { 'persondata.0.isVisible': true },
           ],


EvianTan:
Tested and fixed the bug.
With the role as "Manager" and visibility is "invisible" (image 1), I can see myself on the leaderboard (image 2). Hours shows correctly on summary bar (image 3)
image

image
image
I also use my another "Volunteer" role in the same team to test the invisibility of the "Yiyun Manager", it's not here as expected. (image 4)
image

EvianTan (approved):

PR #337, opened 2023-04-29
Update env file to store the default password

ayush0904:
Hello @EvianTan, As discussed on the slack, can you please verify this. Thank you

# Description
The current implementation of our codebase has a significant security vulnerability, as passwords are hardcoded in multiple locations throughout the code. This means that anyone with access to the repository can easily access these sensitive credentials. To address this issue, we need to implement a more secure approach for storing these passwords. One potential solution is to utilize a separate .env file to store all password information. By doing so, we can ensure that the passwords are kept confidential and are not visible to unauthorized individuals. This approach can significantly reduce the risk of security breaches and protect our system from potential attacks. To implement this solution, we will need to refactor the existing codebase to remove all hardcoded passwords and replace them with calls to the .env file. We will also need to ensure that the .env file is securely stored and that access to it is limited only to authorized individuals. Additionally, we may need to update any associated documentation or processes to reflect this change in our security approach.
Or Implements # (WBS) 

## Related PRS (if any):
No related PR 
…

## Mainly changes explained:
Updated the .env file to inclure the password. (DEF_PWD=123Welcome!)
Made the changes in the places where the password is hardcoded.
…

## How to test:
check into current branch
do `npm install` and `...` to run this PR locally
Check if the login is successgul from different users.

### DIFF
--- a/src/controllers/logincontroller.js
+++ b/src/controllers/logincontroller.js
@@ -11,7 +11,7 @@ const logincontroller = function () {
     const _email = req.body.email;
     const _password = req.body.password;
     
-    //const _defPwd = '123Welcome!';
+    
     const _defPwd = DEF_PWD;
     if (!_email || !_password) {
       res.status(400).send({ error: 'Invalid request' });
--- a/src/controllers/logincontroller.js
+++ b/src/controllers/logincontroller.js
@@ -5,12 +5,14 @@ const config = require('../config');
 const userprofile = require('../models/userProfile');
 
 const logincontroller = function () {
-  const { JWT_SECRET } = config;
+  const { JWT_SECRET,DEF_PWD } = config;
 
   const login = async function _login(req, res) {
     const _email = req.body.email;
     const _password = req.body.password;
-    const _defPwd = '123Welcome!';
+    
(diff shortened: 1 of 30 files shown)


src/controllers/logincontroller.js
```diff
@@ -5,12 +5,14 @@ const config = require('../config');
 const userprofile = require('../models/userProfile');
 
 const logincontroller = function () {
-  const { JWT_SECRET } = config;
+  const { JWT_SECRET,DEF_PWD } = config;
```
EvianTan, line 8: There should be a space between comma and DEF_PWD.

EvianTan:
Hey @ayush0904 ,

Thank you so much for pointing out the problem and raise this PR. Below are things I'd like to discuss with you or like you to address:
1. Other places hard coded with `123Welcome!`
image
A) Can we also use `env.DEF_PWD` in `db.js` instead?
B) How to also hide the default password from frontend `UserProfileAdd.jsx`?

2. Please raise your future PRs use the below template
image

3. Please follow "One PR one (part of) feature/bug" rule, don't include any other changes not related the current PR.

EvianTan (approved):

PR #355, opened 2023-05-17
Fix un-escaped regex

nathanah:
# The Problem
Several checks for the uniqueness of new entries (profiles, badges, etc.) use regexes for case-insensitivity, however, the input was not escaped, so symbols with different meanings in regex were not being treated as text. This makes it so unique, valid emails (and other inputs) were not allowed or caused errors due to “duplication”. For example email@email.com == .....@email.com == e.+@email.com etc.

# Description
This fix escapes regex symbols to instead be treated as text.

There are 6 areas where this change affects. 4 are checking for the uniqueness of items as they are created (user emails for new profiles, badges, projects, and inventory items). The other two are matching email for login and email/name for the forgot password page.

# How to test
uniqueness: Create an item (user profile, badge, project, and inventory item), then attempt to create a separate item with a regex that should match (ex: '123' == '.+'). Order matters.
login: Once you create a new user profile using a regex, make sure you can log into it. 
forgot password: Make sure you can't trigger a password reset email from a regex that could match. 
image
vs
image

### DIFF
--- a/src/utilities/escapeRegex.js
+++ b/src/utilities/escapeRegex.js
@@ -1,6 +1,6 @@
 
 const escapeRegex = function (text) {
-    return text.replace(/[-[\]{}()*+?.,\\^$|#\s]/g, '\\$&');
+    return text.replace(/[[\]{}()*+?.\\^$|]/g, '\\$&');
 };
 
 module.exports = escapeRegex;
--- a/src/utilities/escapeRegex.js
+++ b/src/utilities/escapeRegex.js
@@ -1,0 +1,6 @@
+
+const escapeRegex = function (text) {
+    return text.replace(/[-[\]{}()*+?.,\\^$|#\s]/g, '\\$&');
+};
+
+module.exports = escapeRegex;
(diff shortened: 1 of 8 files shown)


one-community:
Moving straight to final review, thank you Nathan.

20chen-7 (commented):
The function named 'escapeRegex' is clear and accurate which provides reusability for overall application design.
Consider adding some edge cases to validate the input to ensure this function handles any potential cases and behaves as expected.
Overall, the code demonstrates a straightforward approach to escaping special characters for the given text, but further improvements can be made to enhance robustness and reliability.

src/utilities/escapeRegex.js
Cirakuze, line 1: +1 to the other comment. Please consider adding unit testing for `escapeRegex`.

nathanah: I added unit tests. Let me know if you think I missed any edge cases.

20chen-7 (approved):
The updated regular expression has removed redundant specific characters and the unit tests are sufficient for all edge cases based on the regular expression.

Cirakuze (approved):

PR #363, opened 2023-05-26
Bada - include user field isActive in getprojectMembership response

Cirakuze:
Needed for the [FE PR #852](https://github.com/OneCommunityGlobal/HighestGoodNetworkApp/pull/852).

Include the `isActive` user field in the `getprojectMembership` response.


lchua2314 (approved):
Tested the feature and it works as intended. Backend and frontend are both approved. Good job! 

Before:
After

After:
Before

NidaZaki (approved):
Good job! Changes look good to me. Approved the FE as well.

PR #368, opened 2023-06-02
Edwin - Added 3rd toggle to bio announcement

EdwinLauM:
### Description
- Added a third state for the “Bio Posted/Requested” selection to the Weekly Summary Report Page. It now includes "Posted, "Default", "Requested"
- A toggle switch that can change the status of bio announcement has been modified to toggle between three states instead of 2. 

### Mainly changes explained:
Changed the `bioPosted`property from boolean to string to allow for three different states.  

### How to test:
**Please test with the front end [PR 866](https://github.com/OneCommunityGlobal/HighestGoodNetworkApp/pull/866)**
checkout this branch
for the server:
npm run build
npm start
for the front end:
npm run start:local
run the front end and test

PR #399, opened 2023-06-09
Revert "Edwin - Added 3rd toggle to bio announcement"

EdwinLauM:
Reverts OneCommunityGlobal/HGNRest#368

Reverting this PR since it is causing a bug with creating new users. Found the bug in userProfileController.jsx line 188. There is a typo. Instead of `up.permissions = req.body.bioPosted || 'default';` it should be `up.bioPosted = req.body.bioPosted || 'default';`


one-community (approved):
Reverted for this reason:

PR #408, opened 2023-06-16
Lucas fix users own tasks

Lucas-E:
# Description
Users that didn't have the permission to see all the dashboard needed to be in a team to see their own tasks.

## Main changes explained:
When a user is not on a team, only the user's data is fetched, but it didn't include the **role** field, and the role field is necessary so the render function can work, so the **role** property was added when fetching a single user's data.

## How to test:
1. check into current branch
3. Loggin as roles that can't see the full dashboard information, make sure that the user is not on a team
4. check if the user can see itself under the tasks tab


one-community (approved):
Merging without review, Lucas approved :)

PR #427, opened 2023-06-29
Update timerController and re-add oldTimer model

suigunfjc:
# Description

In current dev and prod, the timer will automatically pause every minutes and won't save the status after refreshing the page. 
This PR will fix the current timer in both dev and prod.

The bug is caused by an incomplete revert of the new Timer. When we are currently using the old timer, the controller is trying to store and update data with the `newTimers` doc in the database, which is supposed to be used by the new timer. The incompatible between the frontend and backed causes this bug.

## Main changes explained:
Re-add `oldTimer` in models to connect with `timers` doc in mongodb.
Change the model import in timerController.

## How to test:
1. check into current branch, rebuild and start the HGNRest
2. Test if now the timer is fixed: start/pause/stop/refresh the page, make sure everything is working correctly and the timer won't automatically stop after 1 min.

### DIFF
--- a/src/controllers/REAL_TIME_timerController.js
+++ b/src/controllers/REAL_TIME_timerController.js
@@ -1,6 +1,6 @@
 
 const logger = require('../startup/logger');
-const OldTimer = require('../models/timer');
+const OldTimer = require('../models/oldTimer');
 
 const timerController = function (Timer) {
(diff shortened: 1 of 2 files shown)


one-community (approved):
🔥 fix

PR #432, opened 2023-07-02
Jinchao fix error when editing time logs

suigunfjc:
# Description
Fixed bug:
image

## Main changes explained:
Currently, a time entry can be logged under either a task or a project, and they share a same field in time entry class: projectId
image
The bug is caused by a `throw error` when trying to do `task,findById` with a project id. This PR change it back to a `console.log` and improve the logic of increasing/decreasing timelogged.

## How to test:
1. check into current branch, do build and start.
2. log as admin user
3. Edit the time Entry logged under project or task to see if it can be correctly save now
4. Check if the logged time under a task change correctly when editing a relative time entry.

## Note
We need to refresh the page to load the latest logged time, may need another PR to address that issue.


one-community (approved):
🔥 fix

PR #605, opened 2023-11-11
Aish_Create buildingTools.js

Aishwaryak01:
Created model file for buildingTools.js which consists of schema to hold various attribute information related to tools.

# Description
Created model file for buildingTools.js which consists of schema to hold various attribute information related to tools.
WBS # 4.4.1 from Phase 2 WBS document

## Related PRS (if any):
NA
This is a backend PR

## Main changes explained:
- Created a model file buildingTools.js  which consists of a schema to hold various attribute information related to tools.
- The tool's information is stored on an individual basis.

## How to test:
1. check into the current branch
3. Navigate to HGNRest\src\models\bmdashboard\buildingTools.js
4. Check if all tool-related attributes are listed in the schema.

image

## Note:
This is just a draft of the schema. Once the schema is finalized a collection will be made in the MongoDB database.

### DIFF
--- a/src/models/bmdashboard/buildingTool.js
+++ b/src/models/bmdashboard/buildingTool.js
@@ -3,17 +3,14 @@ const mongoose = require('mongoose');
 const { Schema } = mongoose;
 
 const buildingTools = new Schema({
-    inventoryItemId: { type: mongoose.SchemaTypes.ObjectId, ref: 'INVENTORY' }, // later ,INVENTORY needs to be changed as per inventory model
+    itemType: { type: mongoose.SchemaTypes.ObjectId, ref: 'buildingInventoryType' },
     title: { type: String, required: true },
-    image: { type: String, required: true },
     rentedOnDate: { type: Date, required: true },
     rentDuration: { type: Number, required: true }, // This value should be taken in number of days
-    total: {type: Number, required: true},
-    availableCount:{type: Number, required: true},
     logInStatus:{Boolean},
-    condition:{type: String, enum: ['Good', 'Needs Repair', 'Out of Order','Unused'], default: 'Condition'},
+    condition:{type: String, enum: ['Good', 'Needs Repair', 'Out of Order','Unused'], default: 'Good'},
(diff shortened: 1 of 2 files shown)


tdkent (changes requested):
- schema name should be buildingTool since its a singular item and not plural?
- field `inventoryItemId` can be changed to `itemType: { type: mongoose.SchemaTypes.ObjectId, ref: 'buildingInventoryType' }`
- the image field should probably not be required for now, as it will be quite difficult to test
- can you explain what the `total` and `availableCount` fields are for? I'm not sure they're needed if the model is used to track a single item, but I might be missing something.
- Will the default value of the `condition` field cause an error since it's not in the enum? Should the default instead be 'Good'? See also `purchaseStatus`, should that even have a default value?

Aishwaryak01:
> * schema name should be buildingTool since its a singular item and not plural?
> * field `inventoryItemId` can be changed to `itemType: { type: mongoose.SchemaTypes.ObjectId, ref: 'buildingInventoryType' }`
> * the image field should probably not be required for now, as it will be quite difficult to test
> * can you explain what the `total` and `availableCount` fields are for? I'm not sure they're needed if the model is used to track a single item, but I might be missing something.
> * Will the default value of the `condition` field cause an error since it's not in the enum? Should the default instead be 'Good'? See also `purchaseStatus`, should that even have a default value?

@tdkent : 
> schema name should be buildingTool since its a singular item and not plural? - Sure.Made the change

> field inventoryItemId can be changed to itemType: { type: mongoose.SchemaTypes.ObjectId, ref: 'buildingInventoryType' } - Done

> The image field should probably not be required for now, as it will be quite difficult to test - Removed it.

> Can you explain what the total and availableCount fields are for? I'm not sure they're needed if the model is used to track a single item, but I might be missing something. - I thought this schema was for plural tools so thought of adding the total count and how many of them are available to rent. I now get it that this should be taken care of in inventory.

> Will the default value of the condition field cause an error since it's not in the enum? Should the default instead be 'Good'? See also purchaseStatus, should that even have a default value? - Sure.Made changes.

tdkent (approved):

PR #725, opened 2024-01-27
Rhea creating phase 2 - 7.1.2 New Issue routing, controller / 7.2.1 issue schema

RheaWu1212:
# Description
 Implements # (WBS) phase 2 - 7.1.2 New Issue routing, controller / 7.2.1 issue schema

## Main changes explained:
- creating schema
- set up db in mongoDb compass
- creating routing and controller
- testing

## Note:
- main changes will be complete one by one and testing will be updated...


tdkent (changes requested):
I'm not sure why but it looks like you're again reverting the pull_request_template.md file to an older version. You also have a lot of old merges as part of your commit history. Can you walk me through your git process for this branch? I'm curious to know why these changes are showing up.

The **buildingIssue** model itself is off to a good start. There are some specific changes I would consider making:

- It might be worth having two date fields:
   - a date to track when the issue is created that defaults to Date.now()
   - a user-generated date to track when the issue occurred. It should be required but with no default.
- `staff` should be changed to `createdBy` for clarity
- the BM should be able to add any staff members involved in the issue. This could be accomplished with an array of ObjectIds. The field could be called `staffInvolved` or something like that, and should be not be required.
- `issue` should be changed to `description` or `issueText`, or similar. The maxLength should also be increased to at least 500 characters to allow for a longer description. 
- Add a required `issueTitle` string field so the issue can be composed with both a title and description
- `referredLesson`: Will issues be based off lessons, or will it be the other way around? Would it make more sense to link the issue with a specific project?
- add an optional string array field for storing image urls

RheaWu1212:
> I'm not sure why but it looks like you're again reverting the pull_request_template.md file to an older version. You also have a lot of old merges as part of your commit history. Can you walk me through your git process for this branch? I'm curious to know why these changes are showing up.
> 
> I attached 2 pics here how I did for this PR: 
> `git pull origin main` to pull the latest version > `git checkout -b xxx` to create new bra`nch > git merge development` to merge the latest version to my branch > `git status` and `git add xxx` after I done modifying my file > `git commit` and `push` to github and create PR

Screen Shot 2024-02-03 at 09 35 12
Screen Shot 2024-02-03 at 09 36 05

RheaWu1212:
> * `referredLesson`: Will issues be based off lessons, or will it be the other way around? Would it make more sense to link the issue with a specific project?

> 
Screen Shot 2024-02-03 at 17 52 28
> I actually a bit confused about this prompt, can you please explain it to me based on your understanding?

tdkent:
I think the issue is that you are pulling from `main` but merging `development`. That may be why the merge history is showing up in your commit history. `main` is probably going to be behind `dev`. What I would do is:
- `git switch development`
- `git pull development`
- `git switch your-branch`
- `git merge development` 
Then fix any merge conflicts. That should get your branch up to date with dev rather than main.

tdkent:
> > * `referredLesson`: Will issues be based off lessons, or will it be the other way around? Would it make more sense to link the issue with a specific project?
> 
> > 
> 
> Screen Shot 2024-02-03 at 17 52 28 > I actually a bit confused about this prompt, can you please explain it to me based on your understanding?



> > * `referredLesson`: Will issues be based off lessons, or will it be the other way around? Would it make more sense to link the issue with a specific project?
> 
> > 
> 
> Screen Shot 2024-02-03 at 17 52 28 > I actually a bit confused about this prompt, can you please explain it to me based on your understanding?

That's something I copied from the original version of the Phase 2 docs, but what it's saying is that each issue should have a unique ID (which Mongo will give by default). Then, the Lesson Learned model can have a reference field to the corresponding issue. 

For example if the issue is 'River Sand left out over rainy night turned into mud', the corresponding Lesson Learned might be 'Don't leave sand exposed overnight', with a reference to the issue.

So instead of a `referredLesson` field in the Issue model, there should be a `referredIssues` ObjectId array in the Lesson Learned model.

I hope that makes sense?

PR #1149, opened 2024-11-14
Yashwanth_Backend_for_permission_log_table

yashwanth170:
# Description
Bug_description
The permission change log table only logs changes made to role, so I have added additional features to log permission changes made to individual account. Now the table shows latest logs of changes made to both individual and role 

## Related PRS (if any):
This frontend PR is related to the ##2868 Frontend PR.
To test this backend PR you need to checkout the #2868 frontend PR.
…

## Main changes explained:
- Developed a new MongoDB model to store logs of individual permission changes.
- Implemented a new controller class with a POST method to save data into the MongoDB model and a GET API to retrieve logs from both UserPermissionChangeLog and PermissionChangeLog, sorting them accordingly.
…

## How to test:
1. check into current branch, and check into frontend branch Yashwanth_individual_permission_change_show
3. Clear site data/cache
4. log as admin user
5. go to dashboard→ Other Links → Permission Management → Make changes to a role or individual and verify that the table updates accordingly.

## Note:
Test the updated implementation across different user types and let me know if I missed any expected conventions or if you have additional suggestions. I’ve also resolved an existing issue where submitting without adding or removing any permissions no longer logs duplicate entries for both role and user permissions.


src/controllers/logUserPermissionController.js
```diff
@@ -0,0 +1,82 @@
...
+    }
+  };
+
+  const getPermissionChangeLogs = async function (req, res) {
+    try {
+      const userProfile = await UserProfile.findOne({ _id: req.params.userId }).exec();
+
+      if (userProfile) {
+        if (userProfile.role !== 'Owner') {
+          res.status(204).send([]);
```
nathanah, line 45: Is this checking for permission to view the history? I think Admins might also want to see the history? not sure.
Could use ``await hasPermission(req.body.requestor, 'getPermissionHistory')`` or something (could use an existing permission or create a new one). There shouldn't be hardcoded references to specific roles if avoidable.

nathanah: I see that's just copied from `getPermissionChangeLogs`... I guess it is only Owners then. Would've been easier to find if this was modifying that file instead of creating a new file.

nathanah (changes requested):
The POST request should not have it's own API so that the updates are atomic. The GET request should be updating/replacing the original GET request.

nathanah (approved):

nathanah:
Not sure if this should be on the front-end or during logging on the back-end, but individuals should be labeled as "INDIVIDUAL: <Name>" according to instruction C. in your screenshot. Everything else looks good
image

one-community:
Thank you all, merging!

PR #1188, opened 2024-12-21
Anoushka application search

Anoushka21:
# Description
image
- Search API was already implemented in existing code so I worked on 2,3 and 4.

## Related PRS (if any):
This is a backend PR. You can use Postman to test functionality.
…

## Main changes explained:
- Added a summaries API which fetch job summaries with pagination, search, filtering, and sorting when provided a job title or category. The summary API only shows title,category,location,description,datePosted,featured columns
- Added a job suggestions API which gives suggestions for a drop down menu given s job title search query.
- Added a reset Filter API which resets filters and returns all jobs sorted by title, date posted and featured or not

## How to test:
1. check into current branch
2. do `npm install` and `...` to run this PR locally
3. Go to Postman and test out the endpoints
4.  Use GET `http://localhost:4500/api/jobs/summaries?search=Analyst` for testing Summary API
5. Use GET `http://localhost:4500/api/jobs/suggestions?query=an` for testing the Drop down suggestion API
6. Use GET `http://localhost:4500/api/jobs/reset-filters` for reset filter 

## Screenshots or videos of changes:

https://github.com/user-attachments/assets/2e88366f-4ee8-4290-ae4a-cb2df694630a

PR #1195, opened 2025-01-06
Anoushka Job application search backend  fix summary api

Anoushka21:
# Description
397905963-096787fa-5638-4aca-87f6-74d4577cf6f5

## Related PRS (if any):
This is a bug fix for my last PR [1188](https://github.com/OneCommunityGlobal/HGNRest/pull/1188)
This backend is realated to frontend PR [3020](https://github.com/OneCommunityGlobal/HighestGoodNetworkApp/pull/3020)

## Main changes explained:
- added the jobdetails link in return object of the summary api to ensure that the job summaries are clickable on frontend.

## How to test:
Checkout frontend PR for testing changes.

### DIFF
--- a/src/controllers/jobsController.js
+++ b/src/controllers/jobsController.js
@@ -11,9 +11,13 @@ const getJobs = async (req, res) => {
 
     // Build query object
     const query = {};
-    if (search) query.title = { $regex: search, $options: 'i' }; // Case-insensitive search
+    if (search) {
+      query.$or = [
+      {title: { $regex: search, $options: 'i' }}, // Case-insensitive search
+      { description: { $regex: search, $options: 'i' } }
+      ];
+    }
     if (category) query.category = category;
-
     // Fetch total count for pagination metadata
     const totalJobs = await Job.countDocuments(query);
 
@@ -46,10 +50,14 @@ const getJobSummaries = async (req, res) => {
   try {
     const pageNumber = Math.max(1, parseInt(page, 10)); 
     const limitNumber = Math.max(1, parseInt(limit, 10)); 
-    
     // Construct the query object
     const query = {};
-    if (search) query.title = { $regex: search, $options: 'i' };
+    if (search) {
+      query.$or = [
+      {title: { $regex: search, $options: 'i' }}, // Case-insensitive search
+      { description: { $regex: search, $options: 'i' } }
+      ];
+    }
     if (category) query.category = category; 
 
     // Sorting logic
--- a/src/controllers/jobsController.js
+++ b/src/controllers/jobsController.js
@@ -62,7 +62,7 @@ const getJobSummaries = async (req, res) => {
     // Fetch the total number of jobs matching the query for pagination
     const totalJobs = await Job.countDocuments(query);
     const jobs = await Job.find(query)
-      .select('title category location description datePosted featured') 
+      .select('title category location description datePosted featured jobDetailsLink') 
       .sort(sortCriteria) 
       .skip((pageNumber - 1) * limitNumber) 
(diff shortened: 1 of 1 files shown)


kobakvantrishvili (changes requested):
Left the review on [Frontend PR](https://github.com/OneCommunityGlobal/HighestGoodNetworkApp/pull/3020#pullrequestreview-2535601113).

Anoushka21:
Added new commit which does the following-
- searches jobs across title and description in both `getJobs` and `getJobSummaries` controllers

ziddigsm (changes requested):
The search bar doesn't filter the content based on description. Please check. I have attached a video related to this in the frontend PR.

src/controllers/jobsController.js
```diff
@@ -46,10 +50,14 @@ const getJobSummaries = async (req, res) => {
   try {
     const pageNumber = Math.max(1, parseInt(page, 10));
```
honglinchen0524, line 51: These two constants and pagination logic have been created repeatedly in multiple functions. Consider extracting them into utility functions, to increase code clarity.

honglinchen0524 (changes requested):

src/controllers/jobsController.js
```diff
@@ -11,9 +11,13 @@ const getJobs = async (req, res) => {
 
     // Build query object
     const query = {};
-    if (search) query.title = { $regex: search, $options: 'i' }; // Case-insensitive search
+    if (search) {
+      query.$or = [
+      {title: { $regex: search, $options: 'i' }}, // Case-insensitive search
+      { description: { $regex: search, $options: 'i' } }
+      ];
```
honglinchen0524, line 18: Is there a particular reason why you are using regex for searching? Regex can be quite slow when searching on multiple fields. I recommend adding index and use text search:
```suggestion
 // Add index
Job.collection.createIndex({ title: 'text', description: 'text' });

// Use text search
if (search) {
  query.$text = { $search: search };
}
```

src/controllers/jobsController.js
```diff
@@ -46,10 +50,14 @@ const getJobSummaries = async (req, res) => {
   try {
     const pageNumber = Math.max(1, parseInt(page, 10)); 
     const limitNumber = Math.max(1, parseInt(limit, 10));
```
honglinchen0524, line 52: These two constants and the pagination logic have been created repeatedly in multiple functions (in the jobsController.js file). Consider extracting them into utility functions, to increase code clarity.

honglinchen0524 (changes requested):

Eveye2023:
This PR functions as expected. The `jobDetailsLink` field is successfully returned to the frontend for the `jobs/summaries` endpoint. However, as @honglinchen0524 pointed out, there are several duplicate code sections, and it would be better to extract them into a reusable function.

My test results are as follows:

- The job listings are clickable and lead to the correct job link for the summaries function.

Screenshot 2025-01-29 at 10 52 23 PM

Screenshot 2025-01-29 at 10 53 23 PM



- The job listings are clickable but do not lead to the correct job link for the search function. This is because the frontend doesn't use the jobDetailsLink returned from the backend for the search function.

Screenshot 2025-01-29 at 11 00 43 PM

jaissica (changes requested):
When searching for a keyword in the Summary View, the results are correct, but it redirects to the general job postings view instead of staying in Summary View. I have added a video for [frontend PR](https://github.com/OneCommunityGlobal/HighestGoodNetworkApp/pull/3020).

Ghazi1212 (changes requested):
The search bar is filtering content only by title but not by description.

PR #1196, opened 2025-01-08
Zhifan - Create “Interact with Pause User button” permission

mashpotato9:
# Description
Screenshot 2025-01-08 at 12 03 28 AM

## Related PRS (if any):
This backend PR is related to the [#3023](https://github.com/OneCommunityGlobal/HighestGoodNetworkApp/pull/3023) frontend PR.
To test this backend PR you need to checkout the [#3023](https://github.com/OneCommunityGlobal/HighestGoodNetworkApp/pull/3023) frontend PR.
…

## Main changes explained:
- added pause/resume permission for Admin and Owner
- provided ability to access user management page and the pause button to users assigned with this permission
…

## How to test:
1. check into current branch
3. go to frontend repo for further steps

## Note:

### DIFF
--- a/src/controllers/userProfileController.js
+++ b/src/controllers/userProfileController.js
@@ -1348,7 +1348,7 @@ const userProfileController = function (UserProfile, Project) {
 
     cache.removeCache(`user-${userId}`);
 
-    UserProfile.findById(userId, 'isActive')
+    UserProfile.findById(userId, 'isActive email firstName lastName')
     .then((user) => {
       user.set({
         isActive: status,


src/controllers/userProfileController.js
```diff
@@ -1229,7 +1230,9 @@ const userProfileController = function (UserProfile, Project) {
     );
 
     if (
-      !((await hasPermission(req.body.requestor, 'changeUserStatus')) && canEditProtectedAccount)
+      !((await hasPermission(req.body.requestor, 'changeUserStatus') 
+      || await hasPermission(req.body.requestor, 'pauseResumeUser')) 
+      && canEditProtectedAccount)
```
nathanah, line 1235: This would allow 'pauseResumeUser' to also do what 'changeUserStatus' can do (setting final day for users?).
I don't fully understand why these need to be separate permissions, but they should be different. I asked how this should be handled in the bugs doc.

nathanah (changes requested):

src/controllers/userProfileController.js
```diff
@@ -169,7 +169,8 @@ const userProfileController = function (UserProfile, Project) {
   };
 
   const getUserProfiles = async function (req, res) {
-    if (!(await checkPermission(req, 'getUserProfiles'))) {
+    if (!(await checkPermission(req, 'getUserProfiles') ||
+          await checkPermission(req, 'pauseResumeUser'))) {
```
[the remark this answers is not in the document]
mashpotato9: Team leaders should not be able to pause people on their team. Confirmed with Jae that people with permission to pause anyone should have access to the user management page, which fetches all users.

src/controllers/userProfileController.js
mashpotato9: The pauseResumeUser function does not allow users to set the final day because the “Set Final Day” button only appears when the condition `const canEdit = canEditUserProfile || isUserSelf;`is met. Additionally, users cannot set the final day unless they have the changeUserStatus permission.

The changeUserStatus function manages various user status transitions, including deactivating a user, which can occur with or without a time constraint if one is provided. It handles pausing a user, setting the final day, and directly deactivating a user. All these actions request the same endpoint to toggle the user’s status between active and inactive. Access to these different functionalities is differentiated by the frontend. In the original implementation, deactivating a user with a time constraint required multiple permissions.

For potential security concerns, the pause user functionality has been fully isolated from this function.

nathanah: > For potential security concerns, the pause user functionality has been fully isolated from this function.
Fantastic, thank you.

> The pauseResumeUser function does not allow users to set the final day because the “Set Final Day” button only appears when the condition const canEdit = canEditUserProfile || isUserSelf;is met.
I only raised the concern because people can use tools like Postman to circumvent the front-end limitations and manually construct the API requests. Your solution is the perfect fix for this.

nathanah (approved):

PR #1216, opened 2025-02-04
Luis enable both button function and modal

luisarevalo21:
# Description
This is the backend for issuing two warnings at once. 
The post method modified to allow a single warning and two warnings being issued. 

## Related PRS (if any):
This frontend PR is related to the #XXX backend PR.
To test this backend PR you need to checkout the #XXX frontend PR.
…

## Main changes explained:
- The post method was updated to accommodate the new changes of sending a single warning and two warnings. 
- the email tempalte was modified as well, as two new templates would be sent for the new warnings. 

## How to test:
1. check into current branch
2. do `npm run build && npm run start`
3. follow the frontend PR. 

## Note:

### DIFF
--- a/src/models/currentWarnings.js
+++ b/src/models/currentWarnings.js
@@ -6,6 +6,8 @@ const currentWarnings = new Schema({
   warningTitle: { type: String, required: true },
   activeWarning: { type: Boolean, required: true },
   isPermanent: { type: Boolean, required: true },
+  isSpecial: { type: Boolean, required: true },
+  abbreviation: { type: String },
 });
 
 module.exports = mongoose.model('currentWarning', currentWarnings, 'currentWarnings');
--- a/src/controllers/projectController.js
+++ b/src/controllers/projectController.js
@@ -291,8 +291,17 @@ const projectController = function (Project) {
       res.status(400).send('Invalid request');
       return;
     }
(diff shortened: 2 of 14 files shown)


src/models/currentWarnings.js
```diff
@@ -0,0 +1,13 @@
+const mongoose = require('mongoose');
+
+const { Schema } = mongoose;
+
+const currentWarnings = new Schema({
+  warningTitle: { type: String, required: true },
+  activeWarning: { type: Boolean, required: true },
+  isPermanent: { type: Boolean, required: true },
+  isSpecial: { type: Boolean, required: true },
```
Ankuriboh, line 9: I suggest that you set `default` value to `false`, or unset `required` since this is a schema breaking change.

Ankuriboh (changes requested):
Hi, you might want to restart the PR from the latest `development` branch, as I saw that some of the files do not exist in your branch actually exist in `development`. This could be due to us merged some commit that rollbacked lots of files a while ago.

sundarmachani (changes requested):
I noticed that this PR includes formatting changes such as extra spaces, moving code to new lines, etc. While keeping the code well-formatted is important, these changes also affect the commit history, making it harder to track meaningful changes using tools like git blame. Additionally, having too many commits in a single PR can make it harder to review and maintain a clean commit history.

PR #1221, opened 2025-02-07
added new field `removedDefaultPermissions` in userProfile's permissions obj

sri0606:
# Description
Added a new field in userProfile's permissions object.
image

## Related PRS (if any):
This backend PR is realted to [frontend PR3099](https://github.com/OneCommunityGlobal/HighestGoodNetworkApp/pull/3099)
…

## Main changes explained:
- Added `removedDefaultPermissions` field in permissions object to keep track of defualt role permissions that were removed.
…

## How to test:
1. check into current branch
3. follow instruction on frontend.

## Note:
- check into current branch

### DIFF
--- a/src/models/userProfile.js
+++ b/src/models/userProfile.js
@@ -37,6 +37,10 @@ const userProfileSchema = new Schema({
   permissions: {
     frontPermissions: [String],
     backPermissions: [String],
+    removedDefaultPermissions: {
+      front: [String],  // Default permissions removed for this user
+      back: [String]
+    }
   },
   firstName: {
     type: String,
--- a/src/models/userProfile.js
+++ b/src/models/userProfile.js
@@ -37,10 +37,7 @@ const userProfileSchema = new Schema({
   permissions: {
     frontPermissions: [String],
     backPermissions: [String],
-    removedDefaultPermissions: {
-      front: [String],  // Default permissions removed for this user
-      back: [String]
-    }
+    removedDefaultPermissions: [String]
   },
   firstName: {
     type: String,
--- a/src/utilities/permissions.js
+++ b/src/utilities/permissions.js
@@ -5,6 +5,13 @@ const userService = require('../services/userService');
 const Logger = require('../startup/logger');
 const { PROTECTED_EMAIL_ACCOUNT, ALLOWED_EMAIL_ACCOUNT } = require('./constants');
 
+const hasDefaultPermissionRemoved = async (userId, action) =>
+  UserProfile.findById(userId)
+    .select('permissions')
+    .exec()
+    .then(({ permissions }) => permissions.removedDefaultPermissions.includes(action))
+    .catch(false);
+
 const hasRolePermission = async (role, action) =>
   Role.findOne({ roleName: role })
     .exec()
@@ -18,9 +25,13 @@ const hasIndividualPermission = async (userId, action) =>
     .then(({ permissions }) => permissions.frontPermissions.includes(action))
     .catch(false);
 
-const hasPermission = async (requestor, action) =>
-  (await hasRolePermission(requestor.role, action)) ||
-  hasIndividualPermission(requestor.requestorId, action);
+const hasPermission = async (requestor, action) => {
+  const defaultRemoved = await hasDefaultPermissionRemoved(requestor.requestorId, action);
+  const roleHasPermission = await hasRolePermission(requestor.role, action);
+  const individualHasPermission = await hasIndividualPermission(requestor.requestorId, action);
+
+  return (!defaultRemoved && roleHasPermission) || individualHasPermission;
+}
 
 function getDistinct(arr1, arr2) {
   // Merge arrays and reduce to distinct elements


Peter-Nie2003 (approved):
Work as expected! Well Done!

src/models/userProfile.js
```diff
@@ -37,6 +37,10 @@ const userProfileSchema = new Schema({
   permissions: {
     frontPermissions: [String],
     backPermissions: [String],
+    removedDefaultPermissions: {
+      front: [String],  // Default permissions removed for this user
+      back: [String]
+    }
```
nathanah, line 43:
```suggestion
    removedDefaultPermissions: [String]
```
There's only one list of permissions now, so this should be sufficient. `backPermissions: [String]` in the line above is from before and I didn't want to change the DB schema and have to update every entry.

nathanah (changes requested):
No changes to `hasPermission()` to actually enforce these removed permissions from not being used. 
No changes to `putUser()` (??) to actually save the `removedDefaultPermissions`.

src/models/userProfile.js
sri0606: I initially thought to make the `removedDefaultPermissions` just a list as well, but didn't because if, in future, differentiating between back and front removed default perms is needed, we don't have to change schema and update entries. If `backPermissions` is never used, it does't make sense to have `removedDefaultPermissions.back`. Should I just remove it?

sri0606:
> No changes to `hasPermission()` to actually enforce these removed permissions from not being used. Need something like this in `hasPermission()` (front-end too):
> 
> ```js
> rolePerms = rolePermissions.filter(perm => !removedDefaultPermissions.contains(perm))
> ```
> 
> No changes to `putUser()` (??) to actually save the `removedDefaultPermissions`.

Also, `putUser` already has support for saving `permissions` obj, why do we need to specifically save `removedDefaultPermissions`? 

regarding `hasPermission()`, I've explained a [bit here in this comment](https://github.com/OneCommunityGlobal/HighestGoodNetworkApp/pull/3099#issuecomment-2661231823)

src/models/userProfile.js
nathanah: Correct, removing `removedDefaultPermissions.back` is simpler.

nathanah:
> Also, `putUser` already has support for saving `permissions` obj, why do we need to specifically save `removedDefaultPermissions`? 

Ahh, that makes sense, however, that will give this to everyone that can edit user permissions (I think you had a hard-coded check for Owners on the FE). Not sure if this is core to editing user permissions or if you want to specifically limit it to Owners, but I would lean towards purely relying on permissions instead of hard-coding roles.

sri0606:
> > Also, `putUser` already has support for saving `permissions` obj, why do we need to specifically save `removedDefaultPermissions`?
> 
> Ahh, that makes sense, however, that will give this to everyone that can edit user permissions (I think you had a hard-coded check for Owners on the FE). Not sure if this is core to editing user permissions or if you want to specifically limit it to Owners, but I would lean towards purely relying on permissions instead of hard-coding roles.

In the issue description, Jae mentioned `Owner` roles should be able to manage all permissions. Are you suggesting to add a new permission like 'canEditDefaultPermissions`?

nathanah:
> In the issue description, Jae mentioned `Owner` roles should be able to manage all permissions. Are you suggesting to add a new permission like `canEditDefaultPermissions`?

Either that or add it to `putUserProfilePermissions` (not sure of only Owners have that or if Admins do as well).

sri0606:
> > In the issue description, Jae mentioned `Owner` roles should be able to manage all permissions. Are you suggesting to add a new permission like `canEditDefaultPermissions`?
> 
> Either that or add it to `putUserProfilePermissions` (not sure of only Owners have that or if Admins do as well). Currently, it would be controlled by `putUserProfilePermissions` in `putUser()` if the permissions object is saved together, so checking for Owner specifically in the FE might be redundant.

I have made the necessary updates for `hasPermission()` on both BE/FE. But `putUserProfilePermissions` perm is enabled for both admin and Owner. So, I'm still doing a hard check on FE. Does adding a new perm to Owner role make any changes to entries in DB?

nathanah:
> Does adding a new perm to Owner role make any changes to entries in DB?

Adding permissions to `createInitialPermissions.js` will re-add the listed permissions for each role every time the server restarts, yes.

nathanah (approved):
LGTM

rohith1667 (approved):
checked frontend pr 3099. all the features worked as mentioned. user can add/delete and the features work well in dark mode.

https://github.com/user-attachments/assets/60928754-faaa-4c3b-8531-db3505def26d

Screenshot 2025-03-01 at 5 26 45 PM

SSP24SCM16G (approved):
The PR successfully adds the removedDefaultPermissions field to the userProfile permissions object, enhancing permission tracking. Everything functions correctly, with no issues found

Screenshot 2025-03-01 at 19 02 09
Screenshot 2025-03-01 at 19 02 27

one-community:
Thank you all, merging!

PR #1363, opened 2025-05-03
Humemah: fix - prevent deactivated managers from receiving notifications

humemah:
# Description
image

## Related PRS (if any):
Not related 
…

## Main changes explained:
Main change was to add a condition to stop deactivated from geting emails in task controllers.js
…

## How to test:
1. check into current branch
2. do `npm install` and `...` to run this PR locally
3. Log into the frontend as an owner.
4. Create a test manager account and a test user account, and assign both to the same team.
5. Trigger the sendReviewReq API (e.g., by requesting a task review).
6. 
99279833-8e79-4179-92a9-e9b65b0ea113

7.Check the console to confirm the manager's email appears in the recipients list or that an email is sent to them

## Note:
Include the information the reviewers need to know.

PR #1655, opened 2025-08-13
Humemah Fix deactivated managers still getting notices for blue square reasons being set 

humemah:
# Description
This PR ensures deactivated managers no longer receive notification emails when Blue Square reasons are set.

Previously, PR #1363 addressed most scenarios where deactivated managers would receive these notifications, but a fringe case in the TimeOffRequestController.js file was missed. This fix adds the missing condition to exclude inactive managers from the notification process.

## Related PRS (if any):
This frontend PR is related to the #XXX backend PR.
To test this backend PR you need to checkout the #XXX frontend PR.
…

## Main changes explained:
Update TimeOffRequestController.js to add a condition preventing inactive managers from being included in the notification recipients list.

## How to test:
1. check into current branch
3. Clear site data/cache
4. log as admin user
5. Trigger a Blue Square reason being set for a request where one of the assigned managers is deactivated.
Verify that the deactivated manager does not receive an email notification.

image

## Note:


humemah:
> I tried to verify the feature by:
> 
> 1. Creating a user account with 'Manager' role.
> 2. Added volunteers to that manager account.
> 3. Triggered the 'Schedule Blue Square Reason' from the volunteer account.
> 
> Before deactivating the manager account, I wanted to see if the email reaches the manager but I don't see any emails. Please let me know if this is how the existing feature works.
> 
> After getting the email, I can deactivate the manager account and see if the email is disabled or not.

Thanks for reviewing, @sumedhkumar96!
I tested the full flow on my side and confirmed through console.log that the email trigger executes correctly . it runs for active managers and skips deactivated managers. Everything looks good from my end.

PR #1342, opened 2025-04-17
Reina create interactive map showing location of projects backend

ReinaT5678:
# Description
Backend for frontend PR# ___
Please checkout frontend PR for more details.

## Related PRS (if any):
To test this backend PR you need to checkout the #XXX frontend PR.
…

## Main changes explained:
- Created a new Schema for organization location details 
- Made new routes for retrieving org data from the database
…

## How to test:
1. check into current branch
2. do `npm install` and `npm start` to run this PR locally
3. Check out frontend for further steps

PR #1810, opened 2025-10-15
Purav taking over for Chieh Jui Lee - Fix the conflict for merging pr1342 

JERRRRY:
This is a new PR covering the feedback and solving conflicts on PR1342
https://github.com/OneCommunityGlobal/HGNRest/pull/1342


Anusha-Gali (changes requested):
Hi Chieh Jui Lee,

I have reviewed your PR locally twice and was not able to see the dots in either environments.
1 - Using https://github.com/OneCommunityGlobal/HighestGoodNetworkApp/pull/3432 as Frontend and your PR as backend. I was not able to see any project interactive dots. Nor was i able to load the http://localhost:5173/bmdashboard/totalconstructionsummary (PS. The localhost 3000 doesn't work)
Screenshot 2025-11-14 at 4 40 58 PM

Screenshot 2025-11-14 at 4 41 07 PM

2. Using development as Frontend and your PR as backend. I was not able to load the http://localhost:5173/bmdashboard/interactivemap as well as when i do load  http://localhost:5173/bmdashboard/totalconstructionsummary - I was not able to see any project interactive dots. And the dark mode doesn't let me see any titles
Screenshot 2025-11-14 at 4 46 59 PM

Screenshot 2025-11-14 at 4 49 41 PM

Screenshot 2025-11-14 at 4 48 26 PM

kunchalasireesha (commented):
Hi, it looks like this PR currently has merge conflicts with the base branch. Could you please rebase or merge the latest changes from development into your branch and resolve the conflicts? Once the build passes, I’ll finish the review.

Jaden300 (changes requested):
Image

Image

Image

Image

Image

Image

Image

The interactive map does not fully load and remains in a loading state indefinitely, indicating a likely issue with the underlying data fetch or endpoint response. The total construction summary page loads partially, but several sections are not rendering data as expected. In particular, Financial Loss Tracking, Global Distribution, and Project Status Overview appear blank despite the rest of the layout rendering correctly.

Additionally, there are dark mode inconsistencies where certain card elements that display colors correctly in light mode do not retain their intended styling in dark mode, resulting in missing or inconsistent visual indicators. Overall, core layout functionality is present, but there are data binding issues in multiple sections and theme-related styling inconsistencies that need to be addressed before merge.

CarolineYXW (changes requested):
Hi! I wasn't able to complete testing because the paired frontend/backend branches could not be started successfully.

- The frontend PR (#3432) fails to compile due to case-sensitive import path issues (e.g. constants/lbdashboard vs constants/lbDashboard).
- The backend PR (#1810) could not start because required runtime dependencies were missing after a standard npm install.

Could you please update/rebase the branches and make sure both projects can build successfully before testing?

AnshShah-18 (changes requested):
I reviewed and tested this PR locally.

The backend installed and started successfully, and I verified the following organization-location API behavior:

* Requests without authentication correctly return `401 Unauthorized`.
* `GET /api/bm/orgLocation` returns `200 OK` with six organizations.
* `GET /api/bm/orgLocation/PRJ001` returns the correct organization details.
* A nonexistent organization correctly returns `404 Not Found`.
* The start-date and end-date filters return a filtered response successfully.

However, I found the following blockers:

1. The PR currently has merge conflicts with the `development` branch.
2. In `src/startup/routes.js`, this route was added:

`app.use('api', bmIssueRouter);`

This appears to be a typo and an unnecessary duplicate because the correct route already exists immediately below:

`app.use('/api/bm', bmIssueRouter);`

3. I could not complete frontend integration testing using the paired frontend PR #3432. That branch requires Node 14, while the current project environment uses Node 20, causing dependency installation to fail before the frontend can start.

Please resolve the merge conflicts, remove or correct the duplicate route registration, and provide an updated compatible frontend branch or testing instructions so the complete interactive-map functionality can be verified.

iAbhi001 (changes requested):
I am able to start backend and frontend branches successfully, but none of the wanted resources are found, so requesting a fix, I am attaching the screenshots of what I saw on my screen.

image

Succesful Running through Terminal:

image

image

Please let me know if I am missing anything.

src/controllers/bmdashboard/bmOrgsController.js
```diff
@@ -0,0 +1,96 @@
...
+    const getAllOrgs = async(req, res) => {
+        try {
+            // query parameters             
+            const {startDate, endDate} = req.query;
+            let query = {};
+            
+            // build query if date filters provided             
+            if (startDate || endDate) {
+                if (startDate) query.startDate = {$gte: new Date(startDate)};
+                if (endDate) query.endDate = {$lte: new Date(endDate)};
```
AdiDubbs, line 14: These conditions only return organizations whose entire lifetime sits inside the selected range. An organization that started before startDate but is still active during the range is excluded.

The map uses overlap semantics: endDate >= selected start and startDate <= selected end. Could the API maybe use the same logic?

src/controllers/bmdashboard/bmOrgsController.js
```diff
@@ -0,0 +1,96 @@
...
+            }
+            
+            // execute query and find org
+            const orgs = await Organizations.find(query)
+                .select('orgId name location status startDate country')
```
AdiDubbs, line 19: The Interactive Map uses endDate to determine whether an organization overlaps the selected range and to render its details. This endpoint strips endDate from every response, so the UI falls back to startDate (InteractiveMap.jsx line 349) and excludes organizations that began before the range but are still active during it.

This might also explain the "no dots on the map" reports from @Anusha-Gali back in November and @iAbhi001 last week - with endDate missing, every org collapses to a single-day range in the map's filter, so nothing matches the selected window.

Please include endDate in the selected and transformed fields.

src/controllers/bmdashboard/bmOrgsController.js
```diff
@@ -0,0 +1,96 @@
...
+                success: false,
+                error: 'Server error ' + err.message
+            });
+        }
+    };
+
+    const getOrgById = async (req, res) => {
+        try {
+            const {id} = req.params;
+            Organizations
```
AdiDubbs, line 55: This query is not awaited or returned, so if it rejects the surrounding try/catch never runs. The request can end in an unhandled rejection instead of the expected 500 response. getAllOrgs above already uses await; please use the same pattern here.

The coordinate access below also needs a guard. getAllOrgs filters out records without two coordinates, but this handler reads coordinates[1] directly.

AdiDubbs (changes requested):
The endpoint shape is a good start, and the GeoJSON point model is appropriate for the map.

I am requesting changes because the list response omits endDate even though the Interactive Map uses it for overlap filtering and display; the server-side date filters also exclude organizations active within a selected range; and getOrgById does not await its query or guard malformed coordinates.

I couldn't run the full integration flow because the branch conflicts with development and is 1,814 commits behind it. This needs a rebase that preserves the current routes before the map flow can be retested.

sitaram1921 (changes requested):
Tested locally on `test-merge-1342` as Admin against the dev database, using Postman.
  The API works as documented: unauthenticated requests return 401, `GET /api/bm/orgLocation` returns 200 with six organizations, `GET /api/bm/orgLocation/PRJ001` returns the correct record, a nonexistent ID returns 404, and the startDate and endDate filters both return correctly filtered results. 
The main issue I found is that an invalid date query parameter
  returns a 500 instead of a 400 and leaks the internal error to the client:
  `GET /api/bm/orgLocation?startDate=banana` responds with `Server error Cast to date failed for value "Invalid Date" (type Date) at path "startDate" for model "OrgLocation"`, which exposes the model name and schema paths. The same happens with an invalid `endDate`, so both params should be validated before the query runs. A few smaller things alongside it: `src/startup/routes.js` still contains `app.use('api', bmIssueRouter);`, which is missing its leading slash and duplicates the registration two lines below (@AnshShah-18 raised this on 2026-07-25); `getOrgById` is async and wrapped in try/catch but uses `.then()` without `await` or `.catch()`, so a database error would escape as an unhandled rejection rather than returning a 500, and it reads `org.location.coordinates[1]` unguarded where
  `getAllOrgs` filters for valid coordinates first; `getOrgById` also omits `country` while
  `getAllOrgs` returns it; and `mongoose` is imported in the controller but never used.
  Lastly, the paired frontend PR #3432 is closed, so this API currently has no consumer and the interactive map cannot be verified end to end. Could you confirm which frontend branch this should be tested against? The branch is also conflicting with `development` and the last commit was 2025-10-10, so it will need a merge.

Image
Image
Image
Image

sawantgayatri19 (changes requested):
Tested locally on backend branch: test-merge-1342
Note: End-to-end UI testing could not be completed because the paired frontend PR (#3432) is closed and has Node version dependency blockers. Testing was restricted to API endpoint verification.

Working:

Unauthenticated requests correctly return 401 Unauthorized.

GET /api/bm/orgLocation successfully returns a 200 OK with the organization list.

GET /api/bm/orgLocation/:id successfully returns the correct organization details for valid IDs, and 404 Not Found for nonexistent IDs.

Issues & Blockers Confirmed:

Duplicate Routing: In src/startup/routes.js, there is an unnecessary duplicate and typo: app.use('api', bmIssueRouter); right above the correct /api/bm route.

Unhandled Promises / Server Crashes: In bmOrgsController.js, getOrgById uses .then() but is missing an await or .catch(). If the database query fails, this results in an unhandled promise rejection instead of gracefully returning a 500 error.

Missing Data: The endpoint strips endDate from the list response, which the frontend needs for overlap filtering on the interactive map. country is also omitted from getOrgById.

Lack of Input Validation: Passing an invalid date (e.g., startDate=banana) crashes the endpoint and returns a 500 Server Error that leaks the internal Mongoose schema path to the client instead of properly returning a 400 Bad Request.

Merge Conflicts: There are active conflicts in bmOrgsController.js, orgLocation.js, and routes.js that must be resolved.

one-community:
PR replace with this one: https://github.com/OneCommunityGlobal/HGNRest/pull/2272

PR #1193, opened 2025-01-03
Humera - Taking over for Summit - Fix personal max record badge

skaush21:
# Description

Screen Shot 2025-01-02 at 3 45 53 PM

Fixes:
1. Based on the current design documentation and schema, updated the Data Source to merge Time Entries and savedTangibleHrs to get an accurate reading on the personal best max hour.
2. Modified a date format that fixed the earned dates that were not being added. 

## Related PRS (if any):
Related backend PR: [PR #947 ](https://github.com/OneCommunityGlobal/HGNRest/pull/947)
To test this PR use frontend development.

## Main changes explained:
- Fixed formatting issue, that was causing earnedDates to be inaccurate.
- optimized code, improved clarity, and reduced overhead calls.
-  added a merge function to merge `Time Entries` and `user.savedTangibleHrs`
- cleaned up the code from debugging and uncommented the functioning code.
- Fix new users not being assigned the badge, refactored code and improved added comments for readability
- Fix badgeOfType was being assigned the badgeCollection unnecessarily in the loop that checks for duplicate badges
- removed redundant promise query's

## How to test:
1. check into the current branch
4. do `npm install` and `...` to run this PR locally
5. Clear site data/cache
6. Test 1:
   a.  log in as Owner user
   b. go to dashboard→ Add Time via `Add intangible Time Entry` 
        Note: when logging hours, mark inTangible time as Tangible 
   c. Use Postman to initiate a request.
   d. Navigate to dashboard -> view Profile -> Featured Badges
   e. Verify that the New Personal Max Record Badge icon contains the current 
       personal max record
    f. verify that the badge is assigned in the badgeCollection.
Note: earnedDate will showcase the current date (Sunday) and not the previous dates, this is because `user.savedTangibleHrs` does not have any dates associated with its hours, whereas `Time Entries` does. Due to this, previous dates are not accounted for. 
Walkthrough:

https://github.com/user-attachments/assets/4292fefd-d309-4290-bae5-3dd87d374028

7. Test 2 - MongoDB:
   a.  On MongoDB navigate to userProfile -> savedTangibleHrs. 
   b. Add hours under savedTangibleHrs field.
       Note: The personal best max hour takes the max hour overall weeks. Any hour added must be higher than any 
       previous week. Also, you will need to update lastweektangiblehrs manually in order to test this (should be the 
       previous hour)
   c. Use Postman to send requests and verify in the frontend that the personal best max hour has been updated with the 
       new hour.
[...]

PR #2116, opened 2026-03-22
Sohail: Phase 1 - Fix Personal Max Record badge

sohailuddinsyed:
# Description

This PR fixes critical issues with the Personal Max badge assignment and update logic that were causing the badge to never/incorrectly update when a user broke their personal weekly hours record.

**Fixes:** Personal Max badge assignment and update logic (Priority: High)

Original PR: #1193 

## Issues Fixed:

1. The badge count was being incremented each time a record was broken, when only one Personal Max badge should ever exist per user.
2. A race condition caused `personalBestMaxHrs` to be updated by `$max` earlier in the weekly flow before `checkPersonalMax` ran, meaning the comparison `lastWeek > personalBest` always evaluated to false and the badge never updated.
3. The `earnedDate` was being appended to rather than replaced when a new record was set.
4. The broken condition `lastSaved > lastWeek` was always false since both values are set to the same `timeSpent` in the same atomic DB operation.

## Main changes explained:

- **Updated `src/helpers/userHelper.js`**: Rewrote `checkPersonalMax` function with correct badge assignment logic
  - User now has exactly ONE Personal Max badge at all times
  - When a new weekly record is broken, `earnedDate` is replaced with the current date and `personalBestMaxHrs` is updated in the same atomic DB operation
  - Record comparison now uses `Math.max(...savedTangibleHrs.slice(0, -1))` to derive the previous best from history, bypassing the race condition with `personalBestMaxHrs`
  - A tie does not count as a new record (strictly greater than)
  - Duplicate badge cleanup retained
  - Badge is automatically added for users who do not yet have one
- **Added `src/helpers/__tests__/checkPersonalMax.spec.js`**: Unit test file covering 12 edge cases, all passing

## How to test:

1. Check out current branch
2. Run `npm install` to ensure dependencies are up to date
3. Run `npm run dev` to start the server locally
4. Send POST request to `/api/login`, save the response auth token as a header in Postman
5. Log in as an owner user
6. Connect with our MongoDB server -> UserProfiles Collection -> Your user profile -> Test with different hours in `savedTangibleHrs` (For assuming hours for the most recent week, assign `lastWeekTangibleHours`)
8. Test badge assignment via POST request to `/api/badge/awardNewBadges`

> [!IMPORTANT]
> Replace `awardNewBadges` in `src/helpers/userHelper.js` with this before sending an API call to ensure you only test for your dev account:

```js
const awardNewBadges = async () => {
  try {
    const users = await userProfile
      .find({ isActive: true, firstName: 'Sohail', lastName: 'Admin' })
      .populate('badgeCollection.badge');
    const user = users[0];
    const { _id, badgeCollection } = user;
    const personId = mongoose.Types.ObjectId(_id);
    await checkPersonalMax(personId, user, badgeCollection);
    if (cache.hasCache(`user-${_id}`)) {
      cache.removeCache(`user-${_id}`);
    }
  } catch (err) {
    console.log(err);
    logger.logException(err);
  }
};
```

9. To run the unit tests:

```bash
npx jest src/helpers/__tests__/checkPersonalMax.spec.js --no-coverage --forceExit
```

10. Verify the following scenarios:

### Test Scenario 1: New User — First Week Logging Hours
- Set `savedTangibleHrs` = [10], `lastWeekTangibleHrs` = 10, `personalBestMaxHrs` = 0
- Badge should be added, `earnedDate` = today, `personalBestMaxHrs` = 10

### Test Scenario 2: Record Broken
- Set `savedTangibleHrs` = [10, 8, 15, 12, 20], `lastWeekTangibleHrs` = 20
- Badge `earnedDate` should update to today, `personalBestMaxHrs` = 20

### Test Scenario 3: Record Not Broken / Tied
- Set `lastWeekTangibleHrs` to a value equal to or less than the max of prior weeks
- Badge should remain unchanged

### Test Scenario 4: Zero Hours Logged
- Set `lastWeekTangibleHrs` = 0
- Badge should not update

### Test Scenario 5: Duplicate Badge Cleanup
- If user has multiple Personal Max badges, all but one should be removed automatically

## Expected Results:

- ✅ Each user has exactly ONE Personal Max badge
- ✅ Badge `earnedDate` is replaced (not appended) when a new record is set
- ✅ `personalBestMaxHrs` is updated atomically alongside the badge
- ✅ A tie does not trigger a badge update
- ✅ Zero hours logged does not trigger a badge update
- ✅ New users get the badge assigned on their first week with hours
- ✅ Duplicate badges are cleaned up automatically

## Note:

- The Personal Max badge tracks the highest tangible hours a user has logged in a single week
- The weekly cron job (`awardNewBadges`) runs every Sunday at 12 AM and calls `checkPersonalMax` for each active user
- The record comparison uses `savedTangibleHrs` history (last 200 weeks) excluding the current week, not the stored `personalBestMaxHrs` field, to avoid the race condition with the `$max` update earlier in the weekly flow

### DIFF
--- a/src/controllers/badgeController.js
+++ b/src/controllers/badgeController.js
@@ -3,7 +3,7 @@ const UserProfile = require('../models/userProfile');
 const helper = require('../utilities/permissions');
 const escapeRegex = require('../utilities/escapeRegex');
 const cacheClosure = require('../utilities/nodeCache');
-// const userHelper = require('../helpers/userHelper')();
+const userHelper = require('../helpers/userHelper')();
 
 const badgeController = function (Badge) {
   /**
@@ -13,10 +13,10 @@ const badgeController = function (Badge) {
    */
   const cache = cacheClosure();
 
-  // const awardBadgesTest = async function (req, res) {
-  //   await userHelper.awardNewBadges();
-  //   res.status(200).send('Badges awarded');
-  // };
+  const awardNewBadges = async function (req, res) {
+    await userHelper.awardNewBadges();
+    res.status(200).send('Badges awarded');
+  };
 
   const getAllBadges = async function (req, res) {
     // console.log(req.body.requestor);  // Retain logging from development branch for debugging
@@ -341,7 +341,7 @@ const badgeController = function (Badge) {
   };
 
   return {
-    // awardBadgesTest,
+    awardNewBadges,
     getAllBadges,
     assignBadges,
     postBadge,
--- a/src/helpers/__tests__/checkPersonalMax.spec.js
+++ b/src/helpers/__tests__/checkPersonalMax.spec.js
@@ -1,0 +1,247 @@
+const mongoose = require('mongoose');
+
+/* =======================
+   MOCKS (MUST COME FIRST)
+   ======================= */
+
+jest.mock('../../models/userProfile', () => ({
+  updateOne: jest.fn(),
+  findByIdAndUpdate: jest.fn(),
+}));
+
+jest.mock('../../models/badge', () => ({
+  find: jest.fn(),
+}));
+
+// Mock all heavy dependencies userHelper pulls in
+jest.mock('../../models/team', () => ({}));
+jest.mock('../../models/timeentry', () => ({}));
+jest.mock('../../models/timeOffRequest', () => ({}));
+jest.mock('../../models/profileInitialSetupToken', () => ({}));
+jest.mock('../../models/BlueSquareEmailAssignment', () => ({}));
+jest.mock('../../helpers/dashboardhelper', () => () => ({}));
+jest.mock('../../helpers/helperModels/myTeam', () => ({}));
+jest.mock('../../utilities/emailSender', () => ({}));
+jest.mock('../../utilities/timeUtils', () => ({}));
+jest.mock('../../services/notificationService', () => ({}));
+jest.mock('../../constants/message', () => ({ NEW_USER_BLUE_SQUARE_NOTIFICATION_MESSAGE: '' }));
+jest.mock('../../utilities/nodeCache', () => () => ({
+  hasCache: jest.fn(() => false),
+  removeCache: jest.fn(),
+  getCache: jest.fn(),
+  setCache: jest.fn(),
+}));
+jest.mock('../../startup/logger', () => ({
+  logException: jest.fn(),
+}));
+jest.mock('puppeteer', () => ({}));
+jest.mock('sharp', () => ({}));
+
+/* =======================
+   IMPORTS AFTER MOCKS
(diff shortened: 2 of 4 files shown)


Anusha-Gali (changes requested):
Hi Sohail,

I tried to review your PR locally and i was able to perform the first test sucessfully however for the second test when i have updated the lastWeekTangibleHrs = 20, and ran the postman and the npm test, the personalBestMaxHrs did not update to 20 from 10. As you cam see in the below test 2 screenshots, i had provided the before test run and after test run mongo db screenshots, even though the batches got assigned, the value did not change in personalBestMaxHrs.

**Test 1**
Image
Image
Image
Image
Image
Image

**Test 2**
Image
Image
Image
Image
Image
Image

sohailuddinsyed:
> Hi Sohail,
> 
> I tried to review your PR locally and i was able to perform the first test sucessfully however for the second test when i have updated the lastWeekTangibleHrs = 20, and ran the postman and the npm test, the personalBestMaxHrs did not update to 20 from 10. As you cam see in the below test 2 screenshots, i had provided the before test run and after test run mongo db screenshots, even though the batches got assigned, the value did not change in personalBestMaxHrs.

Hi @Anusha-Gali , 
Thank you for the review. It looks like the issue you encountered is due to the manually assigned database state in Test Scenario 2.

In my logic, the record comparison uses `Math.max(...savedTangibleHrs.slice(0, -1))` to derive the previous best. This assumes that the current week's hours (20) are at the **end** of the array (since MongoDB's `$push` appends to the end).

In your screenshot/values, you assigned `SavedTangibleHours = [..., 20, 12, 15, 8, 10].`

1. The `20 `is at the beginning of the array, so `slice(0, -1)` did not remove it.
2. This caused the logic to compare `lastWeek (20)` against a `previousMax `of **20** (from the start of the array).
3. Since `20 `is not strictly greater than `20`, the update for `personalBestMaxHrs `was correctly skipped.

To test this correctly, please ensure that the `savedTangibleHrs `array either does not contain the `20 `yet, OR that the `20 `is the **last element** in the array. This will allow the `slice(0, -1)` logic to correctly identify the previous best (15 in your case) and trigger the update to 20.

Anusha-Gali (approved):
Hi Sohail,

I have reviewed your PR locally and below are my findings. Thank you for the help.

**Before test**
Image
Image

**Scenario 1:**
Image
Image
Image
Image
Image

**Scenario 2**
Image
Image
Image
Image
Image
Image

**Scenario 3**
Image
Image
Image
Image
Image

**Scenario 4**
Image
Image
Image
Image
Image

**Scenario 5**
Image

one-community:
Thank you all, merging!

PR #2176, opened 2026-04-22
Chirag: Update pagination to be applicable only when specified

ChiragBellara:
# Description
New event creation bug

## Related PRS (if any):
This backend PR is related to OneCommunityGlobal/HighestGoodNetworkApp#5182 frontend PR.

## Main changes explained:
- Updated the GET /events to return all events in default scenarios and paginated results when page and limit are specified.

## How to test:
1. check into current branch
2. do `npm install` and `npm run dev` to run this PR locally
3. Clear site data/cache
4. log as admin user
5. using Postman try hitting the `http://localhost:4500/api/events` -> you should see all the events at once
6. again try hitting `http://localhost:4500/api/events?page=1&limit=10` -> you should see only the first 10 events.

GET default scenario
GET with page and limit

### DIFF
--- a/src/controllers/eventController.js
+++ b/src/controllers/eventController.js
@@ -72,10 +72,11 @@ function getPagination(page, limit, total) {
 }
 
 function formatEvent(event, userId) {
+  event.status = updateEventStatus(event);
+
   const eventObj = event.toObject();
   const waitlist = Array.isArray(event.waitlist) ? event.waitlist : [];
 
-  eventObj.status = updateEventStatus(event);
   eventObj.waitlistCount = waitlist.length;
   eventObj.waitlistEnabled = event.currentAttendees >= event.maxAttendees;
 
@@ -88,9 +89,9 @@ function formatEvent(event, userId) {
   return eventObj;
 }
 
-const getEvents = async (req, res) => {
+const getEvents = async function (req, res) {
   try {
-    const { page, limit, type, location, sortBy, userId } = req.query;
+    const { page, limit, type, location, sortBy } = req.query;
 
     validateQuery({ type, location, sortBy });
 
@@ -104,7 +105,7 @@ const getEvents = async (req, res) => {
       .skip(skip)
       .limit(limitNumber);
 
-    const formattedEvents = events.map((event) => formatEvent(event, userId));
+    const formattedEvents = events.map((event) => formatEvent(event, safeQuery.userId));
 
     res.json({
       events: formattedEvents,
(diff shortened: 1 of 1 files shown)


src/controllers/eventController.js
```diff
@@ -10,51 +10,105 @@ const updateEventStatus = (event) => {
...
+  }
+}
 
-    const totalEvents = await Event.countDocuments(query);
-    let events = await Event.find(query)
-      .populate('resources.userID')
-      .sort({ [sortField]: 1 })
-      .skip((pageNumber - 1) * limitNumber)
-      .limit(limitNumber);
+function buildSafeQuery(location, type) {
```
kunchalasireesha, line 31: Since validateQuery is already being called at the start of the controller to check the VALID_TYPES and VALID_LOCATIONS sets, we don't need these long if/else chains. We can simplify this significantly to make it more readable.

something like this would work

function buildSafeQuery(location, type) {
  const query = { isActive: true };
  if (location) query.location = location;
  if (type) query.type = type;
  return query;
}

src/controllers/eventController.js
```diff
@@ -10,51 +10,105 @@ const updateEventStatus = (event) => {
...
+    pageNumber,
+    limitNumber,
+    skip: (pageNumber - 1) * limitNumber,
+  };
+}
 
-        eventObj.userWaitlistPosition = index !== -1 ? index + 1 : null;
-      }
+function formatEvent(event, userId) {
+  event.status = updateEventStatus(event);
```
kunchalasireesha: Converting Mongoose documents to plain objects (POJOs) is safer to avoid accidental validation errors, unexpected middleware triggers, or unintended state modifications.

suggested changes

const eventObj = event.toObject();
eventObj.status = updateEventStatus(event); // Modify the object, not the document

src/controllers/eventController.js
```diff
@@ -10,51 +10,105 @@ const updateEventStatus = (event) => {
...
+    const totalEvents = await Event.countDocuments(safeQuery);
+    const { pageNumber, limitNumber, skip } = getPagination(page, limit, totalEvents);
+
+    const events = await Event.find(safeQuery)
+      .populate('resources.userID')
+      .sort(sortBy ? { [sortBy]: 1 } : {})
+      .skip(skip)
+      .limit(limitNumber);
+
+    const formattedEvents = events.map((event) => formatEvent(event, safeQuery.userId));
```
kunchalasireesha: safeQuery.userId is currently undefined. The userId is never extracted from req.query, so the waitlist position logic in formatEvent will never trigger for the user.

kunchalasireesha (changes requested):
Nice work refactoring the helper functions out of the main controller! It's much cleaner now.

I've left a few comments regarding:

Bug Fix: Passing the userId through so waitlist positions show up.

Cleanup: Simplifying the query builder logic.

Once these are addressed, this looks good to go!

src/controllers/eventController.js
ChiragBellara: Thank you for the suggestion. The ValidateQuery function is only to check if the user inputs are within the expected range of values. We still need the if else condition because if we use `query.type = type`, we get unsafe query creation errors in SonarQube. I had got this error on my initial commit and hence there are the subsequent checkins to fix this specific error.
The idea of having a separate validation function first is so that if the user enters an unexpected value, the error is caught before the query construction even begins. Since I'm using Set, the validation check will happen in O(1) time. If there are any errors, it will avoid the query construction completely thus removing the possibility of any errors caused by cached queries.

src/controllers/eventController.js
ChiragBellara: This point I understand. The formatEvent function was not created by me and hence I wasn't really sure where it is being used on the frontend. However, what you mention is correct and is also the best practice. I will update the function and checkin those changes.

src/controllers/eventController.js
ChiragBellara: For this, as far as I understand, the Events API is not getting called with UserId as a parameter anywhere on the UI. But as it currently stands the userId is not being extracted from the query. I have updated that part and fixed it in the current checkin.

kunchalasireesha (approved):
Screenshot 2026-05-04 at 10 50 16 PM
Screenshot 2026-05-04 at 10 50 10 PM
lgtm

one-community:
Thank you all, merging!

PR #2181, opened 2026-04-25
Roshini Seelamsetty : fix: implement permissions logic for task extension count visibility

roshini1212:
# Description

Screenshot 2026-04-25 at 11 56 24

## Related PRS (if any):
This backend #2181 PR is related to the #5199 PR Frontend.
To test this backend #2181 PR you need to checkout the #5199 PR frontend.
…

## Main changes explained:
- Update taskController.js to check seeNumberOfTimesTimeAdded / viewTaskExtensionCount permission before returning deadlineCount in getTasks, getTaskById, and getTasksForTeamsByUser
- Add canSeeTaskExtensionCount helper that checks the permission via hasPermission
- Add removeTaskExtensionCounts helper that strips deadlineCount from task results when permission is absent
- Update taskController.spec.js to cover the case where deadlineCount is hidden when permission is missing

## How to test:
1. Check into current backend branch
2. Run npm start
3. Clear site data/cache
4. Log in as Admin
5. Go to Dashboard → Tasks tab
6. Verify deadlineCount is visible for Admin/Owner by default
7. Go to Permissions Management and remove seeNumberOfTimesTimeAdded from a user

## Note:
Requires the frontend PR #5199 to be checked out for full end-to-end testing.

### DIFF
--- a/src/controllers/taskController.js
+++ b/src/controllers/taskController.js
@@ -13,7 +13,24 @@ const followUp = require('../models/followUp');
 const logger = require('../startup/logger');
 
 const taskController = function (Task) {
-  const getTasks = (req, res) => {
+  const canSeeTaskExtensionCount = async (requestor) =>
+    (await hasPermission(requestor, 'seeNumberOfTimesTimeAdded')) ||
+    (await hasPermission(requestor, 'viewTaskExtensionCount'));
+
+  const removeTaskExtensionCounts = (value) => {
+    if (Array.isArray(value)) {
+      value.forEach(removeTaskExtensionCounts);
+      return value;
+    }
+    if (value && typeof value === 'object') {
+      if (Object.prototype.hasOwnProperty.call(value, 'deadlineCount')) {
+        delete value.deadlineCount;
+      }
+      Object.values(value).forEach(removeTaskExtensionCounts);
+    }
+    return value;
+  };
+  const getTasks = async (req, res) => {
     const { level } = req.params;
 
     let query = {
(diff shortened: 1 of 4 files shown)


src/controllers/taskController.js
```diff
@@ -13,7 +13,24 @@ const followUp = require('../models/followUp');
 const logger = require('../startup/logger');
 
 const taskController = function (Task) {
-  const getTasks = (req, res) => {
+  const canSeeTaskExtensionCount = async (requestor) =>
+    (await hasPermission(requestor, 'seeNumberOfTimesTimeAdded')) ||
+    (await hasPermission(requestor, 'viewTaskExtensionCount'));
```
nathanah, line 18:
```suggestion
    (await hasPermission(requestor, 'seeNumberOfTimesTimeAdded'));
```
Is `viewTaskExtensionCount` implemented as something else or a different name for what ended up as `seeNumberOfTimesTimeAdded`? I don't see it used anywhere.

nathanah (approved):
Looks good.

sawantgayatri19 (changes requested):
Thanks for this, the permission-gating logic and helper structure look solid. Before I approve, two things:

Following up on nathanah's comment above — can you confirm whether viewTaskExtensionCount is a permission that actually exists/is assignable elsewhere, or should it be removed per the suggested diff? Right now it's dead-code-shaped and could mislead future permission assignments.
Could you take a pass at the SonarCloud duplication flag (6.7%, need ≤3%)? Might be as simple as factoring the per-method permission check into a single shared call site.

Once those are addressed I'm happy to approve — nice test coverage on the hidden-field case.

PR #2306, opened 2026-08-15
Mahathi - fix: return project names for supplier performance Backend

mahathiganimi:
# Description
Screenshot 2026-08-15 at 3 34 43 PM

## Related PRS (if any):
To test this backend PR you need to checkout the [https://github.com/OneCommunityGlobal/HighestGoodNetworkApp/pull/5448](url) frontend PR

## Main changes explained:
1. Updated the Supplier Performance projects endpoint to retrieve the distinct projectId values referenced by Supplier Performance records.
2. Added a lookup against the real Project collection so the endpoint returns project names instead of only project IDs.
3. Kept the existing Supplier Performance date-range filtering logic unchanged after confirming it was functioning correctly.
4. Verified the previous empty recent-date results were caused by stale 2024 development data rather than incorrect backend filtering.

## How to test:
1. check into current branch
2. do npm install and ... to run this PR locally
3. Clear site data/cache
4. log as admin user
5. go to /bmdashboard/totalconstructionsummary
6. Open the tools and equipment tracking section
7. verify that you can see the supplier performance graph working correctly

## Note:
Note that the local pre-commit related-test command may still fail because the unrelated reasonSchedulingController integration tests cannot connect to the local test database; the backend build itself succeeds.

### DIFF
--- a/src/controllers/summaryDashboard/supplierPerformance.js
+++ b/src/controllers/summaryDashboard/supplierPerformance.js
@@ -1,4 +1,5 @@
 const mongoose = require('mongoose');
+const Project = require('../../models/project');
 const SupplierPerformance = require('../../models/summaryDashboard/supplierPerformance');
 const logger = require('../../startup/logger');
 
@@ -66,25 +67,15 @@ const supplierPerformanceController = function () {
    */
   const getProjectsWithSupplierData = async function (req, res) {
     try {
-      console.log('Fetching projects with supplier performance data');
+      const projectIds = await SupplierPerformance.distinct('projectId');
 
-      // Get distinct project IDs from supplier performance records only
-      const projectsWithData = await SupplierPerformance.aggregate([
-        {
-          $group: {
-            _id: '$projectId',
-          },
-        },
-        {
-          $project: {
-            _id: 1,
-            projectId: '$_id',
-          },
-        },
-        { $sort: { _id: 1 } },
-      ]);
+      const projectsWithData = await Project.find({
+        _id: { $in: projectIds },
+      })
+        .select('_id projectName')
+        .sort({ projectName: 1 })
+        .lean();
 
-      console.log(`Found ${projectsWithData.length} projects with supplier data`);
       res.status(200).send(projectsWithData);
     } catch (error) {
       logger.logException(error);


Adit0717 (changes requested):
Tested the PR locally on Postman. Checked out the mentioned branches. 

Tested the "api/suppliers/performance" endpoint with all filter combinations - projectId, startDate, endDate and everything works correctly. Found one minor issue.

Invalid projectId returns a 500 instead of 400 - Not reachable through the current UI, since the dropdown only sends valid project IDs - but the API route itself doesn't validate projectId format before passing it to mongoose.Types.ObjectId(). If it's called directly (e.g. via Postman) with a wrong or invalid ID, ObjectId() throws and it falls to the catch block, returning 500.

It would be more correct to validate projectId and return a 400 with a clear message (e.g. "Invalid projectId").

Image

DeepighaJ (changes requested):
The filters work good except Tools by Availability section which couldn't be verified since frontend issue.
Image
Image
Image
Image
Image

linlin-husky (changes requested):
Hi @mahathiganimi,

Thank you for your work on this PR! I tested the backend endpoints locally via Postman alongside frontend PR #5448.

### Verified Points:
- **Project Lookup (`GET /api/suppliers/projects`)**: Successfully returns `200 OK` with an array of project objects containing proper `projectName` and `_id` pairs (e.g., "Orientation and Initial Setup", "XiaoFei test project 4").

Image

Image


- **Performance Metrics (`GET /api/suppliers/performance`)**: Correctly returns aggregated metrics (e.g., `onTimeDeliveryPercentage`) for `projectId=all` across the specified date ranges.

Image

- **Required Parameter Check**: The endpoint properly returns `400 Bad Request` with `"Missing required query parameters: startDate, endDate"` when dates are omitted.

Image


### Areas for Improvement / Questions:
1. **Handling Invalid `projectId` Format**:
   - When querying `GET /api/suppliers/performance?projectId=invalid_project_123&startDate=1970-01-01&endDate=2026-08-28`, the endpoint returns a `500 Internal Server Error` due to an unhandled CastError.
  
   - Could we add a format validation check (such as `mongoose.Types.ObjectId.isValid(projectId)`) when `projectId !== 'all'` and return a `400 Bad Request` with an informative error message instead?
  
Image

PR #1140, opened 2024-11-01
Ashish Fix Total team report team code fields

AshishNags:
…active

# Description
Screenshot 2024-11-01 160923

## Related PRS (if any):
To test this backend PR you need to checkout the development frontend PR.
…

## Main changes explained:
- Update getAllTeams function in teamController.js to include the most-used team code
…

## How to test:
1. check into current branch
2. do `npm install` and `...` to run this PR locally
3. Clear site data/cache
4. log as admin user
5. go to dashboard→ Reports→ Reports→Teams
6. verify it shows list of active and most used team code in the team code field

## Screenshots or videos of changes:
Screenshot 2024-11-01 155502

PR #2314, opened 2026-08-21
Yuxuan Fix active teams and team code display

CarolineYXW:
# Description

Fixes the Active Teams and Team Code display issue on the Reports page.
image

The backend previously relied on outdated/deprecated team-level team code data. This update derives each team's displayed Team Code from its active members instead.

The Team Code is selected from active members with valid codes using the most frequently occurring code. Invalid, empty, and inactive-member codes are ignored, and ties are resolved deterministically.

This work is a follow-up to PR #1140.

## Related PRS (if any):

To test this backend PR you need to checkout the development frontend PR [#5459](https://github.com/OneCommunityGlobal/HighestGoodNetworkApp/pull/5459).

## Main changes explained:

- Update `getAllTeams` to determine Team Code from active team members.
- Ignore inactive members and invalid/empty team codes when calculating the displayed code.
- Preserve the complete team member list in the API response.
- Return an empty Team Code when no eligible active-member code exists.
- Add test coverage for active/inactive members, invalid codes, missing profiles, ties, and member preservation.

## How to test:

1. Check out this backend branch.
2. Install dependencies and run the backend locally.
3. Check out the corresponding frontend branch:
   `Yuxuan-fix-active-teams-team-code-display`
4. Run the frontend locally and log in as a Dev Admin.
5. Navigate to Reports → Reports → Teams.
6. Verify that Team Code is derived from active members and that inactive-member codes are not used.
7. Verify teams without a valid active-member Team Code display the expected empty/default state.
8. Verify Active/Inactive filtering continues to work correctly.

aed1320f-f0d8-4c95-a703-23882eae63cd

## Note:


DeepighaJ (changes requested):
Tested the PR. Filtering Works fine
Issues– inactive-member Team Codes are still being used. For example, team 1 has test-2 and H-hhcc on members with visible: false, while the only visible: true member has an empty Team Code. However, the team-level teamCode is returned as H-hhcc. Per the requirement, inactive-member codes should be ignored, and since there is no valid active-member Team Code, the expected result should be the empty/default state. The same issue is also present for Test Team1, where the only coded member is visible: false but the team-level code is test-2.

Image
Image

**Issues:**

- The response shows the only member with a Team Code is inactive, yet the team-level Team Code is still "test-2".

Image
Image

- Verified Team Code doesnt comes from active members and not ignores inactive-member codes, since there is no valid Team Code among the active/visible members, the team should have an empty/default Team Code.
Image

CarolineYXW:
Fixed. The Team Code calculation now excludes members with visible: false while preserving the existing global isActive check. I also added a regression test covering the reported case where inactive team memberships have codes but the active member has no valid Team Code.

PR #2347, opened 2026-09-12
Amaresh: Fix N+1 query and new member threshold in Promotion Eligibility endpoint

amaresh2001:
## Description
Fixes two issues in the Promotion Eligibility endpoint (`POST /api/promotion-eligibility`), found while manually E2E-testing:

1. **Catastrophic slowness (20-60+ seconds per load).** `getPromotionEligibilityData` ran a per-user loop (`users.map(async user => ...)` + `Promise.all`) issuing 3 DB round trips per user: a review-count query, a weekly-hours aggregate, and a `findOneAndUpdate` upsert against ~1,655 seeded users. Classic N+1: hundreds of concurrent aggregation pipelines and upserts queued against a single Mongo instance.
2. **"New Member" threshold didn't match spec.** The doc specifies New Members as "joined ≤ 1 week ago," but the code computed `isNewMember` as `< 6 months` old, which is about 24x more permissive than intended, so the New/Existing Member dropdown was showing the wrong members in each group.

Screenshot 2026-09-12 at 1 21 47 PM
Screenshot 2026-09-12 at 1 21 59 PM
Screenshot 2026-09-12 at 1 22 12 PM

## Related PRS (if any):
This backend PR is related to frontend PR [#5546](https://github.com/OneCommunityGlobal/HighestGoodNetworkApp/pull/5546).

## Main changes explained:
- Update `src/controllers/promotionEligibilityController.js` for replacing the per-user N+1 loop with two batched aggregations run once across all users: `getWeeklyHoursByUser` (grouped by `personId` + week) and `getReviewCountsByUser` (grouped by `resources.userID`, deduped to match the original `countDocuments` semantics exactly).
- Update `src/controllers/promotionEligibilityController.js` for replacing N individual `PromotionEligibility.findOneAndUpdate` calls with a single `bulkWrite`, fired without blocking the response (the response body is already fully computed in memory before the write starts).
- Update `src/controllers/promotionEligibilityController.js` for correcting the `isNewMember` threshold from `< 6 months` to `daysSinceCreated <= 7` (1 week), matching the doc spec.

## How to test:
1. Check out the current branch
2. Run `npm install` and `npm run dev` to run this PR locally
4. Clear site data/cache
5. Log in as an admin user (Dev Admin account)
6. Go to PR Dashboard → Promotion Eligibility (`/pr-dashboard/promotion-eligibility`)
7. Open DevTools → Network tab, reload, confirm `POST /api/promotion-eligibility` completes in ~1-2s (not 20-60s) and the table populates correctly
8. Toggle the New Member / Existing Member dropdown; only accounts created in the last 7 days should appear under "New Member"

Screenshot 2026-09-12 at 1 12 23 PM
Screenshot 2026-09-12 at 1 17 31 PM

## Note:

### DIFF
--- a/src/controllers/promotionEligibilityController.js
+++ b/src/controllers/promotionEligibilityController.js
@@ -10,21 +10,56 @@ const promotionEligibilityController = function (
   Task,
   PromotionEligibility,
 ) {
-  const calculateWeeksMetRequirement = async (userId, pledgedHours) => {
-    const weeklyTasks = await TimeEntry.aggregate([
-      { $match: { personId: mongoose.Types.ObjectId(userId), isTangible: true } },
+  // Batched across all users: one aggregate returns, per user, the list of weekly
+  // hour totals for review/PR-tagged tangible time entries. The per-user pledged-hours
+  // threshold varies, so filtering weeks against it happens in JS after this query returns.
+  const getWeeklyHoursByUser = async (userIds) => {
+    const rows = await TimeEntry.aggregate([
+      { $match: { personId: { $in: userIds }, isTangible: true } },
       { $lookup: { from: 'tasks', localField: 'taskId', foreignField: '_id', as: 'taskInfo' } },
       { $unwind: '$taskInfo' },
       { $match: { 'taskInfo.taskName': { $regex: /review|pr/i } } },
       {
         $group: {
-          _id: { $week: { $toDate: '$dateOfWork' } },
+          _id: { personId: '$personId', week: { $week: { $toDate: '$dateOfWork' } } },
           totalHours: { $sum: { $divide: ['$totalSeconds', 3600] } },
         },
       },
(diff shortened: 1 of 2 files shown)


shubhamjakhete (changes requested):
Requesting changes: `PromotionEligibility.bulkWrite()` is not awaited, so the endpoint can return before the records are saved. This can create a race with the promotion flow.

Please await the single `bulkWrite`. It still avoids the N+1 writes while preserving the previous persistence behavior.

amaresh2001:
@shubhamjakhete thank you for your review. 

promoteMembers's update has no upsert, so if it ran before the snapshot finished writing, the promotion would silently never record isPromoted/promotionDate for that user. Fixed by awaiting the single bulkWrite in 0f11d02.

Ready for re-review.

